PR #200, opened 2026-05-26
feat(globaltool): add in-tool build runner (PR-A of bootstrapper deprecation)

ChrisonSimtian:
## Summary

First of three PRs that deprecate the per-repo \`build.cmd\`/\`build.sh\`/\`build.ps1\` bootstrappers in favour of \`Fallout.GlobalTool\` (\`fallout\`) as the canonical entry point for running builds. This one is **purely additive** — no behaviour change for any repo with existing bootstrappers committed.

Plan & rationale: \`/Users/csimon/.claude/plans/linked-frolicking-pine.md\` (local). Three-PR sequence:

- **PR-A (this one)** — Add in-tool build runner. \`fallout <target>\` runs \`dotnet build\` + \`dotnet run --no-build\` directly instead of spawning \`./build.sh\`.
- **PR-B** — Switch this repo to thin shims, delete \`build.cmd\`, update the \`[GitHubActions]\` generator + workflows, update \`:setup\` scaffolder.
- **PR-C** — Docs cleanup (already tracked under #198).

## What's in this PR

**New files:**

- \`src/Fallout.GlobalTool/BuildProjectResolver.cs\` — resolves the build project (\`.csproj\`) from optional \`BuildProjectFile\` field in \`.fallout/parameters.json\`, falling back to convention \`build/_build.csproj\`. Six unit tests cover convention default, explicit override, missing files (both cases), and empty override values.
- \`src/Fallout.GlobalTool/Program.Run.cs\` — new in-tool runner. Resolves \`dotnet\` (PATH first; shim-provisioned \`.fallout/temp/dotnet-{unix,win}\` as fallback), seeds the same env vars the shim sets, then runs \`dotnet build\` + \`dotnet run --project --no-build -- <args>\` as subprocesses with inherited stdio (colours/progress pass through).
- \`tests/Fallout.GlobalTool.Tests/BuildProjectResolverTests.cs\` — 6 new tests, all green.

**Modified:**

- \`src/Fallout.GlobalTool/Program.cs\` — \`Handle\`'s no-\`:\` fallback now calls \`Run(args, rootDirectory, BuildProjectResolver.Resolve(rootDirectory))\` instead of \`Build(buildScript, ...)\`. The \`buildScript == null\` branch of the "Could not find" prompt is removed — we can run without a shim now — but the \`rootDirectory == null\` branch stays for empty repos.

## What this PR does NOT do (so PR-B has something to do)

[...]

PR #202, opened 2026-05-26
fix(tests): add discoverable test to Nuke.Common.Shim.Tests (fixes windows-latest on main)

ChrisonSimtian:
## Summary

Fixes a latent windows-latest validation failure on \`main\`. \`tests/Nuke.Common.Shim.Tests/\` is a compile-only test project by design (see \`SampleConsumerBuild.cs\` header — "If this file compiles, the MVP shim covers the surface it claims to cover. Nothing here is executed."). It had **zero \`[Fact]\` / \`[Theory]\` methods**.

xUnit's \`dotnet test\` treats "no tests discovered" differently across platforms:

- **Linux / macOS** — emits \`No test is available\` as a warning, **exit 0**.
- **Windows** — same warning, **exit 1**.

The \`Test\` target catches that exit code, throws \`ProcessException\`, and the build aborts. Every push to \`main\` has been silently failing the \`windows-latest\` post-merge validation since this project was added — e.g. [run 26443068694](https://github.com/ChrisonSimtian/Fallout/actions/runs/26443068694) from the #197 merge. \`macos-latest\` and \`ubuntu-latest\` (full) both pass cleanly with the same warning.

## Fix

Add \`ShimCompilesTests.cs\` with a single \`[Fact]\` whose body is \`Assert.True(true)\`. Discovery succeeds on all three platforms now. The comment block on the test class explains *why* this trivial assertion exists and what it actually proves:

> Reaching this assertion proves the assembly loaded, which proves SampleConsumerBuild.cs (and therefore the Nuke.* shim it uses) compiled.

\`SampleConsumerBuild.cs\` is untouched — it remains the real test surface of the shim.

## Verification

- Local \`dotnet test tests/Nuke.Common.Shim.Tests/\` — discovers and passes 1 test.
- Post-merge \`windows-latest\` should go green on the next push to \`main\`.

## Refs

- Noticed while debugging GitHub Actions trigger weirdness on PR #200/#201; not related to that work.

### DIFF
--- a/tests/Nuke.Common.Shim.Tests/ShimCompilesTests.cs
+++ b/tests/Nuke.Common.Shim.Tests/ShimCompilesTests.cs
@@ -1,0 +1,29 @@
+// Copyright 2026 Maintainers of Fallout.
+// Originally based on NUKE by Matthias Koch and contributors.
+// Distributed under the MIT License.
+// https://github.com/ChrisonSimtian/Fallout/blob/main/LICENSE
+
+using Xunit;
+
+namespace Nuke.Common.Shim.Tests;
+
+// The whole point of this project is to verify that SampleConsumerBuild.cs
+// compiles against the Nuke.* shim — the build IS the test. But xUnit's
+// `dotnet test` discoverer treats a test project with zero discoverable tests
+// as an error on Windows (exit code 1) while emitting only a warning on
+// Linux/macOS. That platform-split silently fails the post-merge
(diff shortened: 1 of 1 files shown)


ChrisonSimtian:
Surfaced during work on #203 (bootstrapper deprecation) but lands independently — this fixes a pre-existing windows-latest validation failure on `main` that's unrelated to the bootstrapper work.

PR #230, opened 2026-05-27
Add Building section explaining submodule and build commands

dennisdoomen:
## Summary

Adds a **Building** section to `README.md` that explains:

- The repo uses a git submodule (`vendor/vs-solutionpersistence`) that must be initialised before building — without it `Fallout.VisualStudio.SolutionPersistence` fails with a missing `.resx` error.
- How to initialise the submodule on an existing clone (`git submodule update --init`) and how to pull it in one step when cloning fresh (`git clone --recurse-submodules`).
- The `build.ps1` / `build.sh` entry points and the most common targets (`Compile`, `Test`, `--help`).

Also adds **Building** to the Table of Contents.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,9 +67,46 @@ For per-repo manifest pinning (`.config/dotnet-tools.json`), project setup, and 
 ## Table of Contents
 
 - [Elevator Pitch](#elevator-pitch)
+- [Building](#building)
 - [Build Status](#build-status)
 - [Activity](#activity)
 - [Sponsorship](#sponsorship)
+
+## Building
+
+### Prerequisites
+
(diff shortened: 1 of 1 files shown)


ChrisonSimtian:
thanks for adding that in :-) 
Just a headsup: I do tend to work quite a bit with submodules where it makes sense.

This particular one is probably something that should at some point become part of this project because Microsoft wont be maintaining this library in a way its helpful to us. Might as well move forward, write our own SLNX parser and dunno maybe drag along the SLN one until this format dies? (one day .... together with windows forms and VB :D )

ChrisonSimtian (approved):
lgtm

ChrisonSimtian:
I just replaced the submodule with an inhouse solution, so that part can probably drop off

dennisdoomen:
Ah, then we can close this PR

PR #254, opened 2026-05-28
feat(persistence)!: inline vs-solutionpersistence; rename SolutionModel → Solution; namespace cleanup

ChrisonSimtian:
> ⚠️ **Breaking change** — package IDs and namespaces change. Migration paths preserved via the `Nuke.Common.ProjectModel.*` transition shim.

Closes #248. Codefix follow-up tracked in #253.

## Affected surface
| Was | Now |
|---|---|
| `Fallout.SolutionModel` (package, project, assembly) | **`Fallout.Solution`** at `src/Persistence/Fallout.Solution/` |
| `Fallout.VisualStudio.SolutionPersistence` (package, project, assembly) | **`Fallout.Persistence.Solution`** at `src/Persistence/Fallout.Persistence.Solution/` |
| `Fallout.Common.ProjectModel.*` namespace | **`Fallout.Solutions.*`** (plural per BCL convention) |
| `Microsoft.VisualStudio.SolutionPersistence.{Model,Serializer,Utilities,...}` namespace | **`Fallout.Persistence.Solution.{Model,Serializer,Utilities,...}`** |
| `vendor/vs-solutionpersistence/` submodule | **gone** — sources inlined |
| `ChrisonSimtian/vs-solutionpersistence` repo | now unreferenced; archive in #11 cleanup |

## Summary
- **Vendored fork eliminated.** The fork-of-fork-of-Microsoft (`vendor/` submodule pointing at `ChrisonSimtian/vs-solutionpersistence`) is gone. 92 source files inlined into a new `Fallout.Persistence.Solution` project under `src/Persistence/`, owned outright by Fallout. No more submodule + fork-rebase coordination.
- **Onion layering.** `src/Persistence/` is now the layered home for persistence-ring code. Future persistence-related projects go here.
- **Facade renamed and rehoused.** `src/Fallout.SolutionModel/` → `src/Persistence/Fallout.Solution/`. The rebrand-era namespace mismatch (`Fallout.Common.ProjectModel`) is fixed to match the assembly name — `Fallout.Solutions` (plural to avoid `Fallout.Solution.Solution` awkwardness).
[...]

Issue #253, opened 2026-05-28
Update Nuke→Fallout codefix to map Common.ProjectModel → Solutions for v11


ChrisonSimtian:
## Why
The `NukeMigrationCodeFix` in `Fallout.Migrate.Analyzers` rewrites `Nuke` → `Fallout` token-by-token. Post-v11 (after the `vs-solutionpersistence` inlining in #248), this produces stale namespace paths:

| Input (NUKE-era consumer code) | Codefix output today | Correct v11 target |
|---|---|---|
| `using Nuke.Common.ProjectModel;` | `using Fallout.Common.ProjectModel;` ❌ (namespace gone) | `using Fallout.Solutions;` |
| `Nuke.Common.ProjectModel.Solution x;` | `Fallout.Common.ProjectModel.Solution x;` ❌ | `Fallout.Solutions.Solution x;` |

Consumers who run `fallout-migrate` (or the codefix) on pre-rename NUKE code currently get code that doesn't compile against v11. The transition shim (`Nuke.Components` / `Nuke.Common`) does mirror `Fallout.Solutions.*` back to `Nuke.Common.ProjectModel.*`, so consumers can stay on the shim namespace longer — but the migration codefix should target the canonical v11 namespace.

## Action
- [ ] Extend `NukeToFalloutRewriter` in `src/Fallout.Migrate.Analyzers/NukeMigrationCodeFix.cs` to handle the qualified-name case `Nuke.Common.ProjectModel` → `Fallout.Solutions` (visit `QualifiedNameSyntax` or post-process the rewritten tree)
- [ ] Restore the test expectation in `tests/Fallout.Migrate.Analyzers.Tests/RealWorldSmokeTests.cs` to `Fallout.Solutions.Solution` + `using Fallout.Solutions;` (currently reverted to `Fallout.Common.ProjectModel.*` to keep the test green during #248)
- [ ] Update `TargetFalloutStub` in that test file to also expose `Fallout.Solutions.Solution` (the inline stub only defines `Fallout.Common.ProjectModel.Solution`)
- [ ] Audit other rename-era namespace pairs the codefix might miss (anything else that moved out of `Fallout.Common.*` in v11)

## Done when
- A NUKE consumer running the codefix gets v11-correct namespace paths
- `RealWorldSmokeTests.RewritesRealisticPreRenameBuildClass` expectations reflect the v11 target namespaces
- No regression on simple `Nuke` → `Fallout` cases with no ProjectModel/SolutionModel involvement

PR #255, opened 2026-05-28
test(consumers): add NUKE + Fallout consumer compatibility projects

ChrisonSimtian:
Stacked on #254 (base: \`feat/inline-vs-solutionpersistence\`). After #254 merges, this PR re-targets to \`main\` and the diff shrinks to just the consumer-projects content.

The whole point of this PR is to **prove #254 doesn't break consumers**. If these three consumer projects compile against the #254 source tree, the shim has done its job and #254 is safe to merge.

## Summary

Three small build-project consumers under \`tests/Consumers/\`:

- **\`Nuke.Consumer\`** — pre-rename NUKE-shape \`class Build : NukeBuild\`, \`[Solution]\`, \`Target\` via the \`Nuke.Common\` / \`Nuke.Build\` / \`Nuke.Components\` transition shims. Includes a documented \`using Target = Fallout.Common.Target;\` alias because the shim generator skips delegates by C# language limit (SHIM002).
- **\`Fallout.Consumer.Local\`** — direct \`ProjectReference\` to this repo's local source. Tracks HEAD, catches breakage in the current PR.
- **\`Fallout.Consumer.NuGet\`** — \`PackageReference\` to nuget.org, pinned to **11.0.8** (pre-#254 release), with \`ReplacePackageReferences=false\` so it actually resolves against published packages. Catches packaging issues + upgrade-direction breakage when bumped after a release.

All three are in \`fallout.slnx\` — \`dotnet build fallout.slnx\` (the existing CI gate) compiles them. Compile failure ⇒ a consumer-facing API broke.

## Why no runtime smoke tests

Initial design had a \`Consumers.SmokeTests/\` xUnit project that spawned each consumer via \`dotnet run\` and asserted exit code 0. That approach hits a Fallout framework runtime requirement: \`FalloutBuild.cctor()\` enumerates \`Host\` subclasses and reflects for a static \`IsRunningHost\` property — minimal consumers that don't pull in the full Fallout MSBuild props/targets (\`Fallout.Common.props\`, \`Fallout.Common.targets\`) trigger a \`System.ArgumentException: Host type 'Host' defines no property 'IsRunningHost'\` at activation.

Reproducing the full build-app environment for a smoke consumer is meaningfully more setup than the test's marginal value justifies (the compile-time check catches the ~95% case). The README documents this and points at how to revisit if needed.

## Test plan
- [x] \`dotnet build fallout.slnx\` — 0 warnings, 0 errors. All three consumer projects compile against the #254 changes.
- [x] \`dotnet test fallout.slnx --no-build\` — all suites green (StronglyTypedSolutionGenerator verified snapshot regenerated for the +3 new projects).
- [ ] CI green on this PR (confirms above against a clean machine).
- [ ] On any future PR that breaks a consumer-facing API, one of these three projects fails to compile and CI catches it.

## Coordinates with
- **#254** — the vendor-fork-elimination PR this is stacked on. These tests are the proof that #254's shim work preserves the consumer contract.
- **#253** — codefix follow-up. The \`Fallout.Consumer.NuGet\` will need its pin + Build.cs namespaces bumped to use \`Fallout.Solutions\` after the next release ships the new namespace surface, which will exercise the migration path.

### DIFF
--- a/.fallout/build.schema.json
+++ b/.fallout/build.schema.json
@@ -24,24 +24,7 @@
     "ExecutableTarget": {
       "type": "string",
       "enum": [
-        "CheckoutExternalRepositories",
-        "Clean",
-        "Compile",
-        "CreateGitHubRelease",
-        "DeletePackages",
-        "DownloadLicenses",
-        "GeneratePublicApi",
-        "GenerateTools",
-        "Install",
-        "Pack",
-        "Publish",
-        "References",
-        "ReportCoverage",
-        "Restore",
-        "RunTargetInDockerImageTest",
-        "Test",
-        "UpdateContributors",
-        "UpdateStargazers"
+        "Default"
       ]
     },
     "Verbosity": {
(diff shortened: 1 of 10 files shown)


ChrisonSimtian:
Stack ordering was wrong (consumers built on top of the breaking-change PR, so they tested against the new API rather than catching the change). Superseded by a new PR off main with pre-#254 namespaces — see follow-up. #254 will rebase on top of that.

PR #390, opened 2026-06-16
Use different ansi escape code for yellow on gitlab

ITaluone:
Gitlab does not support "24-bit true-color ANSI sequence" so the logo appears white instead of yellow.

I confirm that the pull-request:

- [x] Follows the contribution guidelines
- [x] Is based on my own work
- [x] Is in compliance with my employer

**Merge preference:** ( ) squash (default)  ( ) rebase


ChrisonSimtian (approved):
lgtm thanks for fixing this :-)

Issue #388, opened 2026-06-16
[GitHubActions] generator: control the shell for run steps (`defaults.run.shell` + per-step `shell:`)


avidenic:
**Description**

GitHub-hosted runners default `run:` steps to different shells per OS (`bash` on Linux/macOS, `pwsh` on Windows), which silently changes script semantics (quoting, error handling, `set -e` vs `$ErrorActionPreference`) in cross-platform jobs. The only robust fix is an explicit shell, which Fallout cannot currently emit. `GitHubActionsRunStep.Write` (`src/Fallout.Common/CI/GitHubActions/Configuration/GitHubActionsRunStep.cs:14-44`) writes the `run:` line and an optional `env:` block but never a `shell:` key, and `GitHubActionsStep` (`Configuration/GitHubActionsStep.cs:6-8`) is an empty abstract base with no `Shell` member.

Proposed change: offer two levels (prefer the first as the primary mechanism):

1. `defaults.run.shell` at workflow (and optionally job) level: set the shell once for *every* `run:` step in scope (tool restore, `dotnet fallout`, and any injected steps) via a single emitted `defaults:` block. This is the idiomatic GitHub mechanism and is cleaner than threading `shell:` through every step type. Surface as e.g. `DefaultShell` on `GitHubActionsAttribute`.
2. Per-step `shell:` as an escape hatch: a nullable `Shell` on the generated run step, emitted only when set.

**Usage Example**

```csharp
[GitHubActions(
    "build",
    GitHubActionsImage.WindowsLatest,
    GitHubActionsImage.UbuntuLatest,
    InvokedTargets = [nameof(Compile)],
    DefaultShell = "pwsh")] // pin shell for ALL run steps in the workflow
class Build : NukeBuild { /* ... */ }
```

Resulting YAML:

```yaml
name: build

on: [push]

defaults:
  run:
    shell: pwsh

jobs:
  windows-latest:
    runs-on: windows-latest
    steps:
      - uses: actions/checkout@v6
      - name: 'Setup: .NET SDK'
        uses: actions/setup-dotnet@v4
      - name: 'Restore: dotnet tools'
        run: dotnet tool restore
      - name: 'Run: Compile'
        run: dotnet fallout Compile
```

**Alternative**

No step type exposes a shell field and `GitHubActionsRunStep.Write` hardcodes a `shell:`-free `run:` line, so the only workarounds are hand-editing the YAML (lost on regen) or forking. 

**Could you help with a pull-request?**

Yes.

PR #393, opened 2026-06-17
Fix errors/warnings for pre-existing code smells after reintroducing editorconfig et al

ChrisonSimtian:
Fixes pre-existing C# warnings / code smells surfaced after reintroducing `.editorconfig` (#388).

**Scope narrowed per @dennisdoomen's review** — the two larger, opinion-dependent changes were split out so this PR stays focused and reviewable:
- Member visibility / encapsulation → #405
- `MSBuildSettings` → `MsBuildSettings` rename → separate PR (pending, breaking-change handling TBD)

## What remains here

**Remove unused method parameters and usings across utilities and CLI**
`ControlFlow`, `SchemaUtility`, `CodeGenerator`, `ProcessTasks`, `GitHubActionsAttribute`, and the CLI navigation/setup/update commands all had method signatures declaring parameters that were never read. Removed the dead parameters and one unused `using` import. A stale splash-screen comment is also corrected.

**Introduce `SingleFileSerializerBase` and clean up the persistence serializer**
The generic arity suffix on the base class name was noise (there is only one concrete base), so it is renamed to `SingleFileSerializerBase`. The `await using FileStream` pattern is replaced with synchronous `using` to fix a `netstandard2.0` compile break (`FileStream` is not `IAsyncDisposable` there — CS8417). Also fixed: a namespace mismatch on `ISerializerModelExtension`, two doc-comment typos in `ISolutionSerializer`, and redundant lambda parameter names in `XmlFolder`.

**Make `PropertyInfo` argument optional for tool-option formatters**
Formatter methods were always invoked with two arguments (`value`, `PropertyInfo`), forcing every formatter to declare the `PropertyInfo` parameter even when unused. Removing it in `CorFlagsSettings` (as part of the unused-parameter cleanup above) caused a parameter-count mismatch at runtime. The invoker now matches arguments to the formatter's declared arity, making the `PropertyInfo` parameter genuinely optional without breaking existing two-parameter formatters.

## Verification
Full `dotnet build fallout.slnx` succeeds (0 errors).

### DIFF
--- a/src/Fallout.Tooling/ToolOptions.Arguments.cs
+++ b/src/Fallout.Tooling/ToolOptions.Arguments.cs
@@ -115,7 +115,7 @@ partial class ToolOptions
                 // The PropertyInfo is an optional second argument: formatters that don't need it
                 // can declare just the value parameter. Pass args matching the method's arity so
                 // both `Format(value)` and `Format(value, PropertyInfo)` shapes are supported.
-                object[] formatterArgs = formatterMethod.GetParameters().Length <= 1
+                object[] formatterArgs = formatterMethod.GetParameters().Length == 1
                     ? [objValue]
                     : [objValue, property];
                 value = formatterMethod.GetValue<string>(obj: this, args: formatterArgs);
--- a/src/Fallout.Tooling/ToolOptions.Arguments.cs
+++ b/src/Fallout.Tooling/ToolOptions.Arguments.cs
@@ -112,7 +112,13 @@ partial class ToolOptions
                 var formatterType = attribute.FormatterType ?? GetType();
                 var formatterMethod = formatterType.GetMethod(attribute.FormatterMethod, ReflectionUtility.All);
                 var objValue = type != typeof(object) ? DeserializeWithCoercion(token, type) : NodeToString(token);
-                value = formatterMethod.GetValue<string>(obj: this, args: [objValue, property]);
+                // The PropertyInfo is an optional second argument: formatters that don't need it
+                // can declare just the value parameter. Pass args matching the method's arity so
+                // both `Format(value)` and `Format(value, PropertyInfo)` shapes are supported.
+                object[] formatterArgs = formatterMethod.GetParameters().Length <= 1
+                    ? [objValue]
+                    : [objValue, property];
+                value = formatterMethod.GetValue<string>(obj: this, args: formatterArgs);
             }
             else
             {
--- a/src/Persistence/Fallout.Persistence.Solution/Serializer/ISolutionSerializer.cs
+++ b/src/Persistence/Fallout.Persistence.Solution/Serializer/ISolutionSerializer.cs
@@ -3,7 +3,7 @@
 
 using Fallout.Persistence.Solution.Model;
 
-namespace Fallout.Persistence.Solution;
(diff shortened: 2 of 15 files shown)


build/Build.CI.GitHubActions.cs
```diff
@@ -53,12 +53,12 @@
     OnPullRequestExcludePaths = new[] { "docs/**", ".assets/**", "**/*.md" },
     InvokedTargets = new[] { nameof(ITest.Test), nameof(IPack.Pack) },
     PublishArtifacts = false)]
-partial class Build
+internal partial class Build
```
dennisdoomen: FYI, in Fallout/Nuke builds, I never bothered with the visibility of types and members and stayed the idiomatic approach. Only when I encapsulate targets or support for targets in a NuGet package, I would do that (and in normal production code).

Also, because of these changes, the PR has become too big too review. Would you mind splitting it in two parts (if you want to keep the visibility)?

ChrisonSimtian: its on its way :-) That was me testdriving rider yesterday night, I thought I might as well do them all in one go ;-)

src/Persistence/Fallout.Persistence.Solution/Serializer/ISolutionSerializer.cs
```diff
@@ -3,7 +3,7 @@
 
 using Fallout.Persistence.Solution.Model;
 
-namespace Fallout.Persistence.Solution;
+namespace Fallout.Persistence.Solution.Serializer;
```
dennisdoomen: 🤔 This looks like a breaking change. Are you sure you want to do this?

src/Fallout.Tooling/ToolOptions.Arguments.cs
```diff
@@ -112,7 +112,13 @@ string Parse(JsonNode token, Type type)
                 var formatterType = attribute.FormatterType ?? GetType();
                 var formatterMethod = formatterType.GetMethod(attribute.FormatterMethod, ReflectionUtility.All);
                 var objValue = type != typeof(object) ? DeserializeWithCoercion(token, type) : NodeToString(token);
-                value = formatterMethod.GetValue<string>(obj: this, args: [objValue, property]);
+                // The PropertyInfo is an optional second argument: formatters that don't need it
+                // can declare just the value parameter. Pass args matching the method's arity so
+                // both `Format(value)` and `Format(value, PropertyInfo)` shapes are supported.
+                object[] formatterArgs = formatterMethod.GetParameters().Length <= 1
```
dennisdoomen: ⛏️ (nitpick) Probably cleaner to use `.Length == 1` as a length of 0 wouldn't be valid.

dennisdoomen (dismissed):
A suggestion and a verification

src/Persistence/Fallout.Persistence.Solution/Serializer/ISolutionSerializer.cs
ChrisonSimtian: good point. I reckon since its in that folder something (claude or ide) mustve decided to adjust the namespace. Good catch :-)

dennisdoomen (approved):

PR #405, opened 2026-06-18
Encapsulate build/component members with explicit visibility

ChrisonSimtian:
Split out of #393 per @dennisdoomen's review that the combined PR had grown too large.

## What this does
Makes accessibility explicit on **shipped and tested** surface — no behavioural change — while deliberately **exempting the dogfood `build/` folder** instead of encapsulating it:

- **Components** (`IFalloutBuild`, `ICreateGitHubRelease`, `IHasTwitterCredentials`, `ISignPackages`, `IToolOptionsWithFramework`): drop the redundant `public` from interface members (implicitly public — matches `for_non_interface_members`).
- **Consumer/sample builds under `tests/`** + `DefaultInterfaceExecutionTest`: explicit `internal`/`private`.
- **`build/` folder: exempted from the rule.** A new `[build/**.cs]` section in `.editorconfig` relaxes `dotnet_style_require_accessibility_modifiers`, so build scripts stay terse (convenience over ceremony — Dennis's "never bothered with visibility in builds"). The explicit modifiers an earlier revision added under `build/` are dropped to match.

## Why exempt build/ rather than encapsulate it
The build project is our own dogfood console app, not shipped API. Forcing explicit accessibility there fights the convenience that makes build scripts pleasant to write, and those modifiers "often just get dropped" anyway. Exempting the folder once in `.editorconfig` is more durable than re-adding modifiers commit by commit.

## Verification
Rebased onto `main` (was conflicting). `dotnet build build/_build.csproj` succeeds, 0 errors.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -149,3 +149,15 @@ resharper_redundant_if_else_block_highlighting = none
 
 # If we use explicit property names in anonymous types, we do that on purpose.
 resharper_redundant_anonymous_type_property_name_highlighting = none
+
+####################################################################
+## Dogfood build scripts (build/)
+####################################################################
+
+# The build project is our own dogfood console app, not shipped API. Favour
+# convenience over ceremony there: explicit accessibility modifiers are optional,
+# so build targets/properties/fields can stay terse. Relax the rule that the
+# global [*.cs] section enforces as an error — both Roslyn (IDE0040) and ReSharper
+# read this key, so a silent severity stops either from flagging build scripts.
+[build/**.cs]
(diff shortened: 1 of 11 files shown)


dennisdoomen:
So you decided to add visibility keywords after all? I'm fine with both, but wondering whether you actually considered my arguments.

ChrisonSimtian:
> wondering whether you actually considered my arguments.

ah I see what you mean, i think i was just a bit sleepdrunk when I did that ... /build can/should be excluded. I'll sort this out tomorrow :-) can probably just drop an ignore override inside the build folder

dennisdoomen (approved):

PR #425, opened 2026-06-25
Allow UTF-8 console input

ITaluone:
_Original description_
> There is a problem for user wanting to input non-ascii characters (ex. cyrillic) during parameter prompt

All credits going to: @rus-art Thank you!

Taken from here: https://github.com/nuke-build/nuke/pull/1321

### DIFF
--- a/src/Fallout.Build/Utilities/ConsoleUtility.cs
+++ b/src/Fallout.Build/Utilities/ConsoleUtility.cs
@@ -47,15 +47,13 @@ public class ConsoleUtility
             }
 
             key = Console.ReadKey(intercept: true);
-            if (ConsoleKey.A <= key.Key && key.Key <= ConsoleKey.Z
-                || ConsoleKey.D0 <= key.Key && key.Key <= ConsoleKey.D9
-                || new[] { '.', '/', '\\', '_', '-' }.Any(x => x == key.KeyChar))
+            if (key.IsValidInputKey())
                 input.Append(key.KeyChar);
             else if (key.Key == ConsoleKey.Backspace && input.Length > 0)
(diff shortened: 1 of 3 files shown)


ChrisonSimtian (approved):

ChrisonSimtian (approved):
generally lgtm, might be good to pull this into a `IsValidInputKey()` extension and put some unit tests around it

dennisdoomen (changes requested):
Can you extend the description a bit? What is it solving? Any chance for some test coverage?

ITaluone:
I was just "backporting" a fix from the original repo, which I thought it would be useful*.

Like the original description says: It previously wasn't allowed to have e.g. cyrillic letters as console input.

Nothing more, nothing less :) 

*) I think we should port over fixes propagated by other NUKE contributors to let them know about this one.

ITaluone:
> generally lgtm, might be good to pull this into a `IsValidInputKey()` extension and put some unit tests around it

On to it

ChrisonSimtian:
can we dial this one back a bit, please? Im all for the interface and such but I think this should be a smaller PR just for the UTF support. We can introduce interfaces and refactor this later :-) 
Just make it an extension and give it a simple unit test, we touch it up later ;-)
@ITaluone

ITaluone:
Ok.. but I don't get it.. you asked for unit tests.. and these are the changes to do so?

Edit: done

ChrisonSimtian (approved):
way better :-) This way the blast radius is a bit smaller. And we hold off with the bigger refactoring a bit, I'm cooking something in behind the scenes that might make the refactoring obsolete ;-)

dennisdoomen (approved):

PR #431, opened 2026-06-25
Allow arbitrary extra `with:` inputs on the checkout step of [GitHubActions]

### LINKED ISSUE #389: [GitHubActions] generator: arbitrary extra `with:` inputs on the checkout step
**Description**

Fallout's `GitHubActionsCheckoutStep` (`src/Fallout.Common/CI/GitHubActions/Configuration/GitHubActionsCheckoutStep.cs`) already emits a solid subset of `actions/checkout` inputs: `submodules`, `lfs`, `fetch-depth`, `progress`, `filter`, plus the fork-aware `ref`/`repository` (`CheckoutRef`). This is genuinely good and we don't want to lose it; the gap is narrow.

`GitHubActionsCheckoutStep.Write` only emits those known keys. There is no escape hatch for the many other `actions/checkout` inputs: `token:`, `ssh-key:`, `path:`, `clean:`, `persist-credentials:`, `sparse-checkout:`, `set-safe-directory:`. If you need any of them, the generator simply can't produce the workflow.

Proposed change: add a `string[] CheckoutWith` property (raw `key: value` lines appended verbatim inside the existing `with:` block, after the typed keys), surfaced setter-only on `GitHubActionsAttribute` alongside `CheckoutRef`. Backward compatible (defaults to empty, emits nothing).

> *Optional, same area:* the action version is hardcoded `actions/checkout@v6`. A `CheckoutAction`/`Version` property would close pinning requests too, though not the headline ask.
>
> *Note:* running shell steps right after checkout is deliberately out of scope here; it belongs to a general step-injection mechanism rather than a second injection path bolted onto the checkout step.

**Usage Example**

```csharp
[GitHubActions(
    "build",
    GitHubActionsImage.UbuntuLatest,
    On = [GitHubActionsTrigger.Push],
    InvokedTargets = [nameof(Compile)],
    Submodules = GitHubActionsSubmodules.Recursive, // existing
    FetchDepth = 0,                                  // existing
    Filter = "blob:none",                            // existing
    CheckoutWith =                                   // NEW: escape hatch
    [
        "token: ${{ secrets.CI_PAT }}",
        "path: src",
        "persist-credentials: false",
    ])]
public partial class Build : NukeBuild { /* ... */ }
```

Resulting YAML:

```yaml
steps:
  - uses: actions/checkout@v6
    with:
      submodules: recursive
      fetch-depth: 0
      filter: blob:none
      token: ${{ secrets.CI_PAT }}
      path: src
      persist-credentials: false
```

**Alternative**

Hand-edit the generated YAML (breaks generate-and-check-in) or drop the generator for those workflows. 

**Could you help with a pull-request?**

Yes.

avidenic:
Closes #389.

Adds `string[] CheckoutWith` to `GitHubActionsAttribute` — an escape hatch for `actions/checkout` inputs the typed knobs don't cover (`token`, `ssh-key`, `path`, `clean`, `persist-credentials`, `sparse-checkout`, `set-safe-directory`). Each entry is emitted verbatim inside the checkout step's `with:` block, after the typed keys, in supply order.

```csharp
[GitHubActions("build",
    GitHubActionsImage.UbuntuLatest,
    InvokedTargets = [nameof(Compile)],
    FetchDepth = 0,
    CheckoutWith = ["token: ${{ secrets.CI_PAT }}", "path: src", "persist-credentials: false"])]
```

```yaml
- uses: actions/checkout@v6
  with:
    fetch-depth: 0
    token: ${{ secrets.CI_PAT }}
    path: src
    persist-credentials: false
```

- Typed keys (`submodules`, `lfs`, `fetch-depth`, `progress`, `filter`, `ref`/`repository`) always emit first; `CheckoutWith` lines follow, in order.
- Raw passthrough, no validation — the build author owns the YAML. This also lets multi-line block scalars work (e.g. `sparse-checkout: |` with continuation lines as further entries), which a `KEY: value` validator would reject.
- Defaults to `new string[0]` and emits nothing when unset — existing workflows are byte-for-byte unchanged.

**Scoped out** (per the issue): the `actions/checkout@v6` version override (clean additive follow-up) and post-checkout shell steps (a general step-injection mechanism, not this issue).

**Tests** — three Verify snapshot cases: `checkout-with` (typed key + the issue's example, proving ordering), `checkout-with-only` (guard opens `with:` with no typed key), `checkout-with-sparse` (multi-line block scalar — the case raw passthrough exists to support). No `*ValidationTest.cs`: the feature never throws, and the repo reserves those for validating features.

---


ChrisonSimtian (approved):
lgtm

PR #432, opened 2026-06-25
Remove `IHasChangelog` / `IHazChangelog` build component

ITaluone:
I don't know if this is the correct way of dealing with the deleted changelog.. But worth a try...


src/Fallout.Components/ICreateGitHubRelease.cs
```diff
@@ -42,7 +42,7 @@ async Task<Release> GetOrCreateRelease()
                             Name = Name,
                             Prerelease = Prerelease,
                             Draft = Draft,
-                            Body = ChangelogTasks.ExtractChangelogSectionNotes(ChangelogFile).JoinNewLine()
+                            Body = ""
```
ITaluone, line 45: Here, I don't know what actually should go here.. 🤷‍♂️

dennisdoomen:
I don't think this is the solution. That `IChangeLog` interface is part of the built-in "components". But I'm still trying to understand how it is supposed to be used.

build/Build.cs
```diff
@@ -21,7 +21,6 @@
 [ShutdownDotNetAfterServerBuild]
 partial class Build
     : FalloutBuild,
-        IHasChangelog,
```
ChrisonSimtian, line 24: yes that will unbreak our internal build now that `Changelog.md` doesnt exist anymore :-)

src/Fallout.Components/ICreateGitHubRelease.cs
```diff
@@ -14,7 +14,7 @@
 namespace Fallout.Components;
 
 [ParameterPrefix(GitHubRelease)]
-public interface ICreateGitHubRelease : IHasGitRepository, IHasChangelog
```
ChrisonSimtian, line 17: this we should look into, however this is a breaking change that would break GH releases for people.
Current behaviour seems to be that it does the changelog based on the Changelog.md
Since this is its own mechanism, I'm quite happy to pull this apart at some point. Just not here and now

ChrisonSimtian (commented):
I'm happy for us to look into the `Build.cs` change (although PR #433  will fix the root cause), the rest we should probably park on the side for now.

build/Build.cs
dennisdoomen: No, it won't, as our build implements `ICreateGitHubRelease` and that one inherits `IHasChangeLog`

dennisdoomen:
As discussed, we'll solve this in #433

PR #375, opened 2026-06-06
Add GitHub release notes config and deprecate CHANGELOG.md

dennisdoomen:
Switches release notes from the manually maintained CHANGELOG.md to GitHub's automatic release notes feature.

## What changed

- Added .github/release.yml to configure automatic GitHub release notes, with categories matching the repo's label conventions (reaking-change, enhancement, ug, security, dependencies, documentation, internal/process/
efactor/chore). Dependabot PRs and PRs labelled skip-changelog or dependencies are excluded.
- Marked CHANGELOG.md as deprecated with a notice pointing to .github/release.yml and the GitHub Releases page.
- Added rule #7 to AGENTS.md explicitly banning conventional-commit prefixes on commit messages and PR titles.
- Added matching guidance to CONTRIBUTING.md under the PR-writing section.

Related to #263

PR #433, opened 2026-06-26
Handle missing changelog (sections) during pack

dennisdoomen:
Summary
- Make pack resilient when no changelog release sections are available.
- Add regression coverage for a missing changelog file.
- Clarify that the build entrypoint project is intentionally excluded from solution builds.

Changes
- Handle missing changelog sections during pack
  - Return empty release notes when the changelog file is missing or contains no release sections, avoiding the pack-time exception.
- Adjust solution build settings
  - Keep the solution file aligned with the current build configuration for this PR branch.
  - The build entrypoint project (`build/_build.csproj`) is intentionally marked as not built in the solution so it is exercised via the build entrypoint (`./build.ps1` / `./build.sh`) rather than as a normal solution build target.

Combined effect
- Pack no longer fails just because the changelog is absent or empty, and the PR history is split into focused commits.

Validation
- `dotnet test tests/Fallout.Common.Tests/Fallout.Common.Tests.csproj --filter ChangelogTasksTest`

**Note** The failing build was caused by #375, but wasn't caught by GHA as that one skips changes to doc files.

### DIFF
--- a/src/Fallout.Common/ChangeLog/ChangeLogTasks.cs
+++ b/src/Fallout.Common/ChangeLog/ChangeLogTasks.cs
@@ -12,6 +12,11 @@ using Serilog;
 // ReSharper disable ArgumentsStyleLiteral
 namespace Fallout.Common.ChangeLog;
 
+/// <summary>
+/// Provides a set of tasks and utility methods for working with changelogs in a software repository.
+/// This class includes methods for reading, extracting, and finalizing changelog entries, as well as
+/// preparing release notes for publishing on platforms such as NuGet.
+/// </summary>
 public static class ChangelogTasks
 {
     public static string GetNuGetReleaseNotes(string changelogFile, GitRepository repository = null)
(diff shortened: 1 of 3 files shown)


src/Fallout.Common/ChangeLog/ChangeLogTasks.cs
ChrisonSimtian, line 1: ah yes, thats probably the way. Fault tolerant, if its not there skip.
Mind to call the logger to spit out a warning? I'd expect if someone calls the Changelog Task they intend to write it (which in our case was the actual issue, we didnt fully remove it from our internal build process before I killed the `Changelog.md`)

ChrisonSimtian (approved):
lgtm, I'd throw a warning ala

> Warning: You called the Changelog Task but we couldnt find a changelog.md

And then we can either create one (I'd rather not) or the user at least sees why we didnt write anything

src/Fallout.Common/ChangeLog/ChangeLogTasks.cs
dennisdoomen: I decided to use information logging, as you may use the `ICreateGitHubRelease` interface, but didn't expect to inherit `IHasChangeLog` as well.

dennisdoomen:
> And then we can either create one (I'd rather not) or the user at least sees why we didnt write anything

See my remark about logging.

PR #434, opened 2026-06-27
Skip convention-less Host subclasses in host auto-detection

ChrisonSimtian:
## Problem

`Host.Default` scans every public subclass of `Host` and **asserted** each defines a static `IsRunning{Name}` property. The `Nuke.*` transition shim emits `Nuke.Common.Host` (a public `Host` subclass with no `IsRunningHost`), so detection threw `ArgumentException: Host type 'Host' defines no property 'IsRunningHost'` — inside `FalloutBuild`'s **static constructor**. Net effect: **any build that merely references the shim crashes at startup, before a single target runs.**

## Fix

Treat a `Host` subclass that doesn't follow the `IsRunning{Name}` convention as *not running* instead of throwing. Defensive against any third-party `Host` subclass, not just the shim. Non-breaking, additive behavior.

## Test

Adds a **runtime** regression in `Nuke.Common.Shim.Tests` — the existing `SampleConsumerBuild` shim test is compile-only, which is exactly why this slipped through (it never runs a build). The new test reproduces the precise error without the fix and passes with it.

## How it was found

By the new `fallout.canary` acceptance suite, which consumes the published `-preview` packages and *runs* real consumer builds. See Fallout-build/Fallout.Canary#3.

### DIFF
--- a/src/Fallout.Build/Host.Activation.cs
+++ b/src/Fallout.Build/Host.Activation.cs
@@ -29,9 +29,15 @@ public partial class Host
     private static bool IsRunning(Type hostType)
     {
         var propertyName = $"IsRunning{hostType.Name}";
-        var member = hostType.GetProperty(propertyName, ReflectionUtility.Static)
-            .NotNull($"Host type '{hostType.Name}' defines no property '{propertyName}'");
-        return member.GetValue<bool>();
+        var member = hostType.GetProperty(propertyName, ReflectionUtility.Static);
+
+        // A public Host subclass that doesn't follow the `IsRunning{Name}` convention is
+        // not a runnable host — treat it as not-running rather than throwing. Host
+        // auto-detection runs in FalloutBuild's static ctor, so an exception here crashes
+        // every build that merely references such an assembly. The Nuke.* transition shim
+        // emits exactly such a subclass (`Nuke.Common.Host`), which made any shim-referencing
+        // build abort on startup. See Fallout.Canary#3.
+        return member is not null && member.GetValue<bool>();
     }
 
     private static Host CreateHost(Type hostType)
--- a/tests/Nuke.Common.Shim.Tests/HostDetectionTests.cs
+++ b/tests/Nuke.Common.Shim.Tests/HostDetectionTests.cs
@@ -1,0 +1,34 @@
+using System.Linq;
+using System.Reflection;
+using Xunit;
+
+namespace Nuke.Common.Shim.Tests;
+
+// Runtime counterpart to the compile-only SampleConsumerBuild: actually exercises host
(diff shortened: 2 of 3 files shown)


ChrisonSimtian:
@dennisdoomen small PR to fix something that came back from a new canary repo I created this morning.
Re canary:
basically runs our preview i.e. github release once a night and raises issues in that repo for anything that fails.
I thought that might be a cool way to make sure `main` doesnt just build but also builds (phun intended ;-) )

tests/Nuke.Common.Shim.Tests/HostDetectionTests.cs
```diff
@@ -0,0 +1,34 @@
...
+        // Force the Nuke.Common shim assembly to load so detection sees its types.
+        var shimAssembly = typeof(global::Nuke.Common.NukeBuild).Assembly;
+
+        // Precondition: the shim really does emit a public, convention-less Host subclass.
+        // Without one, this regression test has no teeth.
+        var offenders = shimAssembly.GetTypes()
+            .Where(t => t.IsPublic && t.IsSubclassOf(typeof(global::Fallout.Common.Host)))
+            .Where(t => t.GetProperty($"IsRunning{t.Name}", BindingFlags.Public | BindingFlags.Static) is null)
+            .ToList();
+        Assert.NotEmpty(offenders);
```
dennisdoomen: ⛏️(nitpick) Not using FluentAssertions?

tests/Nuke.Common.Shim.Tests/HostDetectionTests.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        // Touching FalloutBuild runs its static ctor -> Host.Default, scanning every public
+        // Host subclass. Resolving the property (not throwing) is the assertion.
+        var host = global::Fallout.Common.FalloutBuild.Host;
+        Assert.NotNull(host);
```
dennisdoomen: ⛏️Same here

src/Fallout.Build/Host.Activation.cs
```diff
@@ -29,9 +29,15 @@ internal static IEnumerable<Type> AvailableTypes
     private static bool IsRunning(Type hostType)
     {
         var propertyName = $"IsRunning{hostType.Name}";
-        var member = hostType.GetProperty(propertyName, ReflectionUtility.Static)
-            .NotNull($"Host type '{hostType.Name}' defines no property '{propertyName}'");
-        return member.GetValue<bool>();
+        var member = hostType.GetProperty(propertyName, ReflectionUtility.Static);
```
dennisdoomen, line 32: 👍Love the rationale. That will help us in the future.

dennisdoomen (approved):

PR #444, opened 2026-06-29
Refactor the UTF8 console input

ITaluone:
Refactor on top of #425


ChrisonSimtian:
just adding to this: https://github.com/ChrisonSimtian/Fallout/pulls
I'm doing a bigger refactoring of the underlying architecture so this might or might not become obsolete once this lands.
While I do appreciate your work in this area, I would suggest we wait until this piece has landed and then look into it again from a new perspective :-)

ChrisonSimtian:
lets target this for v2026 as soon as we got the next minor release out :-)

ChrisonSimtian (approved):
lets get that in now, small steps but we're going into the right direction 😬 thanks for helping pushing this up the hill 😊

ITaluone:
so... merge?

Issue #315, opened 2026-05-29
[Foundation] Epic: Engine de-statification & extension-pipeline formalization (plugin foundation)


ChrisonSimtian:
## Epic: Engine de-statification & extension-pipeline formalization

Foundation work for the plugin architecture (this milestone). The current engine bootstraps per-run state into process-global statics, walks targets strictly sequentially, and discovers extensions via magic priority numbers with non-deterministic tie-breaking. This epic makes the engine **reentrant, test-isolatable, parallel-ready, and plugin-ready** — entirely via backwards-compatible (facade-preserving) refactors.

All public static build properties (`FalloutBuild.RootDirectory`, `Verbosity`, `Host`, …) are **preserved as facades** over an internal `BuildContext` — no public breakage, so this stays on the 2026 line and does not need the year-cut breaking-change flow.

### Out of scope

- **Caching / incrementality.** Verified there is zero infrastructure for it today (only the coarse `--continue` invocation-hash). It's a large, product-level decision and the "delegate to MSBuild" stance is defensible. Tracked separately if/when desired.

### Tickets & dependency order

- [ ] #306 — FT-1: Fix cross-invocation static-state leaks *(bug, S, no deps — ship first)*
- [ ] #307 — FT-2: Introduce internal `BuildContext` + ambient activation *(L, deps FT-1)*
- [ ] #308 — FT-3: Convert public statics to `BuildContext`-backed facades *(M, deps FT-2)*
- [ ] #309 — FT-4: Instance-based, re-parseable `ParameterService` *(M, deps FT-2)*
- [ ] #310 — FT-5: Context-scope tool-path resolvers + process defaults *(M, deps FT-2)*
- [ ] #311 — FT-6: Thread-safe, per-run-scoped logging *(M, deps FT-2; prereq FT-8)*
- [ ] #312 — FT-7: Formalize the extension pipeline *(M, deps FT-2)*
- [ ] #313 — FT-8: Opt-in DAG-parallel target execution *(L, deps FT-3/4/5/6)*
- [ ] #314 — FT-9: Reentrancy/isolation test harness *(S, deps FT-1..FT-5)*

```
FT-1 ─▶ FT-2 ─┬─▶ FT-3 ─┐
              ├─▶ FT-4 ─┼─▶ FT-8 (parallel exec)
              ├─▶ FT-5 ─┤
              ├─▶ FT-6 ─┘
              └─▶ FT-7
[...]

Issue #306, opened 2026-05-29
[Foundation] FT-1: Fix cross-invocation static-state leaks (event handlers, log sink, verbosity mappings)


ChrisonSimtian:
## Summary

Fix cross-invocation static-state leaks in the engine. These are **latent bugs today** — they manifest whenever `BuildManager.Execute` runs twice in one process (test harnesses, the `dotnet fallout` global tool, future in-process/IDE scenarios).

## Findings (file:line)

- `BuildManager.cs:42` — `ToolOptions.Created += …` subscribed every run, never removed.
- `BuildManager.cs:41` / `:69` — `Console.CancelKeyPress +=` and `CancellationHandler += Finish` accumulate; `s_cancellationHandlers` (`:20`) is never cleared.
- `Logging.cs:214-233` — `InMemorySink.Instance._logEvents` accumulates; `Dispose()` (which clears) is never called → run 2 reports run 1's errors/warnings.
- `VerbosityMappingAttribute.cs:34-46` — `Mappings` is a `LookupTable` (dict-of-lists) appended to on every `OnBuildInitialized` with no dedup → duplicate keys pile up.

## Scope

- Make event subscriptions idempotent (subscribe-once, or unsubscribe in `finally`).
- Clear/reset `InMemorySink` per run (or call `Dispose`).
- Dedup or reset `VerbosityMapping.Mappings` per run.

## Acceptance criteria

- A test that calls `Execute` twice in one process shows: no duplicated handlers firing, no carried-over log events, no duplicate verbosity mappings.
- No public API change.

## Notes

- **Non-breaking.** No dependencies — do this first; it's shippable on its own.
- Size: **S**.

PR #450, opened 2026-06-30
Reset per-run global state at the end of Execute (cross-invocation leaks, FT-1)

ChrisonSimtian:
First step of the engine de-statification ([Foundation] epic [#315](https://github.com/Fallout-build/Fallout/issues/315)) — **FT-1 / [#306](https://github.com/Fallout-build/Fallout/issues/306)**. Non-breaking groundwork before the `BuildContext` foundation (FT-2).

`BuildManager.Execute` leaked process-global state across invocations, so a second build in the same process (tests, hosted/reentrant scenarios) inherited the first's residue:

- `Console.CancelKeyPress` / `ToolOptions.Created` handlers were re-subscribed every call and never removed — **accumulating**;
- the cancellation-handler list kept each run's `Finish` (capturing a stale `build`);
- `Logging.InMemorySink.Instance` carried log events over;
- `ValueInjectionUtility`'s value cache and the NuGet/Npm tool-path resolver config persisted.

**Fix:** the two global handlers are held in locals and the `finally` undoes exactly what the run set — unsubscribes both, drops `Finish` from the cancellation list, and resets the carried-over state (`InMemorySink.Clear`, `ValueInjectionUtility.ClearCache`, `NuGet`/`NpmToolPathResolver.Reset`). `new T()` moved ahead of the subscriptions so a failed build creation leaks nothing.

No public API change. The reentrancy guarantee itself is asserted by FT-9 (the isolation harness); this PR removes the leaks it will assert against. Full build + test suite green.

### DIFF
--- a/src/Fallout.Build/Execution/BuildManager.cs
+++ b/src/Fallout.Build/Execution/BuildManager.cs
@@ -18,12 +18,12 @@ internal static class BuildManager
 {
     private const int ErrorExitCode = -1;
 
-    private static readonly LinkedList<Action> s_cancellationHandlers = new();
+    private static readonly LinkedList<Action> cancellationHandlers = new();
 
     public static event Action CancellationHandler
     {
-        add => s_cancellationHandlers.AddFirst(value);
-        remove => s_cancellationHandlers.Remove(value);
+        add => cancellationHandlers.AddFirst(value);
+        remove => cancellationHandlers.Remove(value);
     }
 
     [ModuleInitializer]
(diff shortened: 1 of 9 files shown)


ITaluone (approved):

ChrisonSimtian:
@dennisdoomen this one please, should be quick :-)

dennisdoomen (approved):
🤔 Can't we add some test to verify this behavior?

PR #455, opened 2026-06-30
Properly implement `FancyNames` feature for solution generator

### LINKED ISSUE #445: [Solution] FancyNames option is dead code
### Problem
The `FancyNames` option for the `[Solution]` strongly-typed accessor does nothing, two ways: `SolutionAttribute` exposes no `FancyNames` property (only `List`/`GenerateProjects`), so `[Solution(FancyNames = true)]` won't compile; and even when forced on, the emitter strips the fancy `٠` separator so output is identical to the default. Surfaced in #442.

### Outcome
`FancyNames` is either a working, settable option or removed entirely — no dangling reads left in the emitter, generator, or console.

### Acceptance criteria
- [ ] `[Solution(FancyNames = true)]` compiles and changes accessor names (`Foo.Bar` → `Foo٠Bar`), **or** the concept is removed across `SolutionEmitter`, `StronglyTypedSolutionGenerator`, and `Fallout.Solution.Codegen`.
- [ ] A spec in `Fallout.Solution.Codegen.Specs` pins the chosen behaviour.

### Notes
- No-op root cause: `GetEscapedName` does `.Replace(".", "٠")` then `.ReplaceRegex(@"[^A-Za-z0-9_]", "_")` (ASCII-only), which deletes the `٠` — `src/Fallout.Solution.Codegen.Emit/SolutionEmitter.cs:91`.
- Attribute missing the property: `src/Fallout.Common/Attributes/SolutionAttribute.cs`. Generator read: `src/Fallout.SourceGenerators/StronglyTypedSolutionGenerator.cs:52`.
- Non-breaking either way.

ITaluone:
Closes: #445


ChrisonSimtian:
there is a bunch of generated files in *.Tests where they'd probably should be in *.Specs

src/Fallout.SourceGenerators/StronglyTypedSolutionGenerator.cs
ChrisonSimtian, line 1: probably doesnt work for the likes of `.github` where the dot is the first character of the name.
Can you add a test for that please and see how it copes? I suspect it breaks in the compiler, because its an invalid char at the start of an identifier. I reckon thats where the whole `00-bla` thing came from

src/Fallout.SourceGenerators/StronglyTypedSolutionGenerator.cs
```diff
@@ -124,9 +124,8 @@ internal class {{ name }}(SolutionFolderModel model, Fallout.Solutions.Solution
                     string GetEscapedName(string name)
                     {
                         name = name
-                            // .Replace(".", fancyNaming ? "丨" : "_")
                             .Replace(".", fancyNaming ? "٠" : "_")
-                            .ReplaceRegex(@"[^A-Za-z0-9_]", _ => "_");
+                            .ReplaceRegex(@"[^A-Za-z0-9_٠]", _ => "_");
```
ChrisonSimtian: can we grab the homoglyph (the `٠`) and make it a constant?
Would be cool if we could just do a drop-in replacement if we feel like there better alternative.

Also, Claude reckons we should use latin Lo (a letter) U+A78F ` ꞏ` because it IS allowed at the beginning of an identifier (its a letter not a digit)
Worth investigating, which makes the unit test around this even more valuable :-)

ChrisonSimtian (dismissed):

src/Fallout.SourceGenerators/StronglyTypedSolutionGenerator.cs
ITaluone: I don't get it. What has the .github directory to do with this generator?

src/Fallout.SourceGenerators/StronglyTypedSolutionGenerator.cs
ITaluone: Replaced every hard-coded `_` with either this fancy char or `_` depending on the flag.. let my know what you think

ITaluone:
> there is a bunch of generated files in *.Tests where they'd probably should be in *.Specs

Ahh yes.. these were the old snapshots

ChrisonSimtian (approved):

PR #464, opened 2026-07-01
Fix `UpdateSolutionFileContentSpecs` not testing anything

ITaluone:
I realized while studying the `UpdateSolutionFileContentSpecs` while preparing #461 that the tests above actually not testing anything, because passing in a string and verify this exact string isn't what a test is for ;)

### DIFF
--- a/tests/Fallout.Cli.Specs/UpdateSolutionFileContentSpecs.cs
+++ b/tests/Fallout.Cli.Specs/UpdateSolutionFileContentSpecs.cs
@@ -37,63 +37,12 @@ public class UpdateSolutionFileContentSpecs
                 		SolutionGuid = {61B2112A-2DA2-45AC-AFF5-6C50A10BB92F}
                 	EndGlobalSection
                 EndGlobal
-                """,
-        """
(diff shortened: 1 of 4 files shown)


tests/Fallout.Cli.Specs/UpdateSolutionFileContentSpecs.cs
```diff
@@ -112,28 +61,15 @@ public class UpdateSolutionFileContentSpecs
...
-    public Task Test(int number, string input, string expected)
+    public Task Test(int number, string input)
     {
         var content = input.SplitLineBreaks().ToList();
         Program.UpdateSolutionFileContent(content, "RELATIVE", "GUID", "NAME");
 
-        return Verifier.Verify(expected)
+        var updatedContent = string.Join(Environment.NewLine, content);
+
+        return Verifier.Verify(updatedContent)
```
ITaluone:
```suggestion
        return Verifier.Verify(string.Join(Environment.NewLine, content))
```

ChrisonSimtian (approved):

PR #469, opened 2026-07-06
Change diagnostic ID from `NUKE001` to `FALLOUT001` in `StrongTypedSolutionGenerator`

ITaluone:
Just fell over this legacy `NUKEXX` diagnostic id... I don't know if this was left behind intentionally, so I added a new one additionally...

### DIFF
--- a/src/Fallout.SourceGenerators/StronglyTypedSolutionGenerator.cs
+++ b/src/Fallout.SourceGenerators/StronglyTypedSolutionGenerator.cs
@@ -138,7 +138,7 @@ public class StronglyTypedSolutionGenerator : ISourceGenerator
         }
         catch (Exception exception)
         {
-            var diagnostic = Diagnostic.Create(
+            var diagnosticLegacy = Diagnostic.Create(
                 "NUKE001",
                 nameof(StronglyTypedSolutionGenerator),
                 exception.Message,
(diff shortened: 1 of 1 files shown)


ChrisonSimtian:
probably just overlooked. Look we do a major version bump soon, I'd just get rid of the legacy one and do a simple rename.
Could be really fancy and make this a constant somewhere, but I wouldnt run both IDs at the same time tbh

ITaluone:
Ok.. so when you are fine by "just" renaming the ID I am fine too

src/Fallout.SourceGenerators/StronglyTypedSolutionGenerator.cs
```diff
@@ -138,15 +138,25 @@ string GetEscapedName(string name)
         }
         catch (Exception exception)
         {
-            var diagnostic = Diagnostic.Create(
+            var diagnosticLegacy = Diagnostic.Create(
```
ITaluone:
```suggestion
            var diagnostic = Diagnostic.Create(
```

src/Fallout.SourceGenerators/StronglyTypedSolutionGenerator.cs
```diff
@@ -138,15 +138,25 @@ string GetEscapedName(string name)
...
                 "NUKE001",
```
ITaluone:
```suggestion
                "FALLOUT001",
```

src/Fallout.SourceGenerators/StronglyTypedSolutionGenerator.cs
```diff
@@ -138,15 +138,25 @@ string GetEscapedName(string name)
...
+            var diagnosticNew = Diagnostic.Create(
+                "FALLOUT001",
+                nameof(StronglyTypedSolutionGenerator),
+                exception.Message,
+                DiagnosticSeverity.Error,
+                DiagnosticSeverity.Error,
+                isEnabledByDefault: true,
+                warningLevel: 0);
+            context.ReportDiagnostic(diagnosticLegacy);
+            context.ReportDiagnostic(diagnosticNew);
```
ITaluone:
```suggestion
            context.ReportDiagnostic(diagnostic);
```

ChrisonSimtian (approved):
yeah im fine with it. That might be somewhat breaking but I'd say its just something we missed in the rebrand, so more a bug that needs fixing because expected behaviour would've been "FALLOUT001" instead of "NUKE001" :-)

Could probably also use some tests at some point but meh, lets keep it simple and small :-)

Issue #336, opened 2026-05-30
[GitHubActions] generator: emit CD pipelines (or a thin shell delegating to dotnet fallout Deploy)


ChrisonSimtian:
Part of #332.

The `[GitHubActions]` generator models "run these targets," not deployment pipelines — so `release.yml`/`experimental.yml`/`preview.yml` are hand-written. Extend it to either:
- emit multi-environment, gated CD pipelines, **or**
- emit a thin generated shell that delegates to `dotnet fallout Deploy` (preferred — keeps the logic in C#, same as validation calls `dotnet fallout Test`).

**Done when** the publish workflows are generated, not hand-authored (modulo the platform approval gate).

PR #475, opened 2026-07-10
Add a typed step-injection pipeline to the GitHub Actions generator

### LINKED ISSUE #456: [GitHubActions] generator: a typed step-injection pipeline (custom steps at named positions)
**Description**

Add one capability to the generator: inject arbitrary workflow steps at chosen positions in a generated job. Today the step sequence is fixed and closed — `GitHubActionsAttribute.GetSteps()` (`src/Fallout.Common/CI/GitHubActions/GitHubActionsAttribute.cs:236-284`) yields checkout → optional cache → the setup-dotnet / tool-restore / `dotnet fallout` run block → optional artifacts, with no hook points. `GitHubActionsStep` (`Configuration/GitHubActionsStep.cs`) is an empty abstract base, and every concrete step pins a hardcoded `uses:` (checkout@v6, setup-dotnet@v4, cache@v4, upload-artifact@v5). There is no way to inject a step that runs a marketplace action (`github/codeql-action/analyze`, `docker/build-push-action`, `aquasecurity/trivy-action`) or a custom shell step. `GetSteps` is `private`, so even subclassing can't reach it.

**Why a typed code hook rather than another attribute property.** The recently merged generator features (`Env`, `RunsOnLabels`, `DefaultShell`, `CheckoutWith`) are all declarative, finite-cardinality scalars, so an attribute property is the right home for each. A step is different in kind: it is a *dictionary* (`with:`, `env:`) **plus ordering plus multiple instances at multiple positions**, none of which a C# attribute can carry — attribute arguments are limited to primitives, strings, enums, `Type`, and arrays of those; never `IDictionary`, objects, or delegates. That ceiling is exactly why a NUKE fork would resort to a reflection + stringly-typed method-name hook, which silently emits an *incomplete* workflow on any typo / wrong signature, is invisible to rename refactors and find-usages, and breaks under trimming/AOT. Fallout owns the `Build` base class, so it has a strictly better, compile-checked seam available. The proposal is three pieces.

**(a) A public, user-constructible step type.**

```csharp
public sealed class GitHubActionsCustomStep : GitHubActionsStep
{
    public string? Name { get; init; }
    public string? Uses { get; init; }                  // null Uses => pure `run:` step
    public IDictionary<string, string> With { get; init; } = new Dictionary<string, string>();
    public IDictionary<string, string> Env  { get; init; } = new Dictionary<string, string>();
    public string? If { get; init; }
    public string? Shell { get; init; }                 // run steps only
    public string[] Run { get; init; } = [];             // single entry => `run: x`; multiple => `run: |` block scalar
    public bool? ContinueOnError { get; init; }
    public int? TimeoutMinutes { get; init; }
    public string? Id { get; init; }
}
```

Validation at generation time — fail the build, never emit invalid YAML silently (consistent with the colon assertions proposed for `Env` and the typed `workflow_dispatch` inputs):

- exactly one of `Uses` or a non-empty `Run` must be set — both, or neither, is a build error;
- a non-empty `With` requires `Uses` (`with:` is only valid on `uses:` steps);
- `Shell` is valid only on a run step (`Uses == null`);
- a run step that sets no `Shell` inherits the workflow-level `defaults.run.shell` (the merged `DefaultShell` feature) automatically — no special-casing.

A single custom step type covers both `uses:` and `run:` steps; a separate script-step type is unnecessary (a `null` `Uses` is the run step).

**(b) Named positions**, anchored to the always-present checkout and run-block so they stay well-defined when cache/artifacts are absent:

```csharp
public enum GitHubActionsStepPosition
{
    PostCheckout,   // after checkout, before cache
    PreRun,         // after cache (if any), before the setup-dotnet / restore / dotnet fallout block
    PostRun,        // after the run block, before the built-in artifact upload
    JobEnd,         // after the built-in artifact upload — end of the job
}
```

**(c) A typed placement hook** — an interface the build implements, invoked **once per generated job**, with a write-only insertion API and read-only access to that job's built-in steps:

```csharp
public interface IConfigureGitHubActions
{
    void ConfigureSteps(GitHubActionsStepPipeline pipeline);
}

public sealed class GitHubActionsStepPipeline
{
    // context — which job is being configured
    public string WorkflowName { get; }
    public GitHubActionsImage Image { get; }

    // read-only — inspect the built-in steps already assembled for this job
    public IReadOnlyList<GitHubActionsStep> BuiltInSteps { get; }

    // write — contribute insertions (no remove, no modify)
    public void Insert(GitHubActionsStepPosition position, GitHubActionsCustomStep step);
    public void Insert(GitHubActionsStepPosition position, IEnumerable<GitHubActionsCustomStep> steps);
}
```

The generator queries `Build as IConfigureGitHubActions` and, for each generated job, calls `ConfigureSteps` with a pipeline scoped to that job, then splices the collected insertions in at their positions (multiple insertions at one position render in call order). Because the hook is invoked **per job** and the pipeline carries the job's identity, one mechanism scopes steps to a workflow (`WorkflowName`) or a runner (`Image`) by ordinary `if` branching — no per-step scoping arrays needed. The pipeline is the only public surface for placement; `BuiltInSteps` is a read-only view (a build inspects it by pattern-matching the concrete step types — checkout, cache, run, artifact — to adapt to a job's shape; the exact richness of that view is a detail we're happy to follow your lead on), so the generator stays in sole control of the base sequence.

*(If you'd rather avoid a new interface, a `protected virtual GetSteps` or `ConfigurePreRunSteps`/`ConfigurePostRunSteps` seams would also work, and we'll PR whichever you prefer — but the interface keeps the generator in control of the base sequence, whereas an overridable `GetSteps` makes user code re-emit and re-splice the full internal sequence and couples it to internal ordering.)*

**Relationship to the multi-job / job-DAG work (#324, #336).** This hook is designed to compose with a future multi-job generator rather than be reworked by it. Two deliberate choices carry it forward: the hook is invoked **per job** (today one job per image; under a job DAG, one call per DAG node), and the job identity lives on the **pipeline context object**, not the interface signature — so that context can later gain a DAG job's key, the target(s) it runs, and its runner without breaking `ConfigureSteps(pipeline)`. Read-only `BuiltInSteps` lets injection adapt to a job's shape (e.g. a job that already contains a `download-artifact` step is a downstream DAG node). The new step types a DAG needs — `download-artifact` and friends — are already expressible here as ordinary `Uses` custom steps. Job-level concerns (`needs:`, job `outputs:`, environment gates) are explicitly **out of scope** for this step pipeline; they belong to the job model in that work. We'd suggest landing this issue first, as the step-injection substrate the DAG work can build on.

**Usage Example**

```csharp
[GitHubActions("security", GitHubActionsImage.UbuntuLatest,
    On = [GitHubActionsTrigger.Push], InvokedTargets = [nameof(Compile)])]
partial class Build : NukeBuild, IConfigureGitHubActions
{
    public void ConfigureSteps(GitHubActionsStepPipeline pipeline)
    {
        if (pipeline.WorkflowName != "security")
            return; // scope these steps to one of several [GitHubActions] workflows

        pipeline.Insert(GitHubActionsStepPosition.PostCheckout, new GitHubActionsCustomStep
        {
            Name = "Setup Node",
            Uses = "actions/setup-node@v4",
            With = new() { ["node-version"] = "20" },
        });

        pipeline.Insert(GitHubActionsStepPosition.PostRun, new GitHubActionsCustomStep
        {
            Name = "Perform CodeQL Analysis",
            Uses = "github/codeql-action/analyze@v3",
            If = "github.ref == 'refs/heads/main'",
        });
    }
}
```

Resulting YAML:

```yaml
steps:
  - uses: actions/checkout@v6
  - name: Setup Node               # injected at PostCheckout
    uses: actions/setup-node@v4
    with:
      node-version: 20
  - uses: actions/cache@v4
    # ...
  - name: 'Setup: .NET SDK'
    uses: actions/setup-dotnet@v4
  - name: 'Restore: dotnet tools'
    run: dotnet tool restore
  - name: 'Run: Compile'
    run: dotnet fallout Compile
  - name: Perform CodeQL Analysis  # injected at PostRun
    uses: github/codeql-action/analyze@v3
    if: github.ref == 'refs/heads/main'
```

**Alternative**

Hand-edit the generated YAML (overwritten on regen), maintain a separate hand-written workflow alongside the generated one, or fork the generator. A NUKE fork can do the last via a reflection-based hook (a custom-steps type plus stringly-typed pre/post method-name properties), but that is exactly the fragile mechanism this issue avoids. This issue upstreams the *capability* with a typed, compile-checked mechanism instead of reflection, folding what would otherwise be two step types into one (a `null` `Uses` is the run step).

**Could you help with a pull-request?**

Yes.

avidenic:
Closes #456.

## Problem
`GitHubActionsAttribute.GetSteps()` yields a fixed, closed step sequence with no hook points and is `private`. No way to inject a marketplace action or custom shell step at a chosen position.

## Outcome
A build implements `IConfigureGitHubActions.ConfigureSteps(GitHubActionsStepPipeline)` to inject typed `uses:`/`run:` steps at four named positions — compile-checked, no reflection.

```csharp
partial class Build : NukeBuild, IConfigureGitHubActions
{
    public void ConfigureSteps(GitHubActionsStepPipeline pipeline)
    {
        if (pipeline.WorkflowName != "security") return;   // scope by workflow / runner
        pipeline.Insert(GitHubActionsStepPosition.PostRun, new GitHubActionsCustomStep
        {
            Name = "Perform CodeQL Analysis",
            Uses = "github/codeql-action/analyze@v3",
            If = "github.ref == 'refs/heads/main'",
        });
    }
}
```

New public types: `GitHubActionsCustomStep`, `GitHubActionsStepPosition` (`PostCheckout`/`PreRun`/`PostRun`/`JobEnd`), `IConfigureGitHubActions`, `GitHubActionsStepPipeline`.

## Notes
- Positions anchor to the always-present checkout + run block, so they resolve whether or not cache/artifacts are present.
- Validation fails the build (never emits invalid YAML): exactly one of `Uses`/`Run`; `With` requires `Uses`; `Shell` only on a run step.
- Additive / non-breaking. Generator stays in sole control of the base sequence; the hook only inserts.
- The hook is invoked **per job** with identity on the pipeline context (not the interface signature) and a read-only `BuiltInSteps` view — deliberately composes with the #336 job-DAG work rather than being reworked by it. Job-level concerns (`needs:`/`outputs:`/env gates) are out of scope for #336.
- New API follows the subsystem's existing idioms; matched to `[GitHubActionsInput]`'s validation style.

## Acceptance criteria
- [x] Inject `uses:`/`run:` at each of the four positions; multiple inserts at one position render in call order.
- [x] Positions resolve correctly with cache/artifacts absent.
- [x] Invalid steps fail generation with a clear `ArgumentException`.
- [x] Steps scope to a workflow/runner via `WorkflowName`/`Image` (no per-step scoping arrays).
- [x] No churn on existing snapshots.

## Verification
Full `Fallout.Common.Specs` suite: 114 passed, 7 skipped, 0 failed. Zero snapshot churn. Full-solution build clean.


tests/Fallout.Common.Specs/CI/GitHubActionsCustomStepSpecs.cs
```diff
@@ -0,0 +1,107 @@
...
+                              Name = "Setup Node",
+                              Uses = "actions/setup-node@v4",
+                              With = new() { ["node-version"] = "20" }
+                          });
+
+        yaml.Should().Be(
+            "- name: Setup Node\n" +
+            "  uses: actions/setup-node@v4\n" +
+            "  with:\n" +
+            "    node-version: 20\n");
```
IT-VBFK:
```suggestion
        yaml.Should().Be(
            """
            - name: Setup Node
              uses: actions/setup-node@v4
              with:
                node-version: 20
            """);
```
I vote for string literal in general, like so

avidenic: Done, tnx for good suggestion. The exact-output asserts  should now be passing on the windows/macos jobs (the PR gate only runs ubuntu) by normalizing line endings in a `ShouldRenderAs` helper. The renderer writes `Environment.NewLine`, so CRLF on Windows, and a raw literal has no trailing newline, so it normalizes both sides to `\n` and adds the terminal one.

IT-VBFK: Hmm I think this line-ending thing is already handled by `""" """`?

ChrisonSimtian:
nice feature, thanks for bringing this in :-)

src/Fallout.Common/CI/GitHubActions/Configuration/GitHubActionsCustomStep.cs
```diff
@@ -0,0 +1,80 @@
...
+        if (ContinueOnError.HasValue)
+            Scalar("continue-on-error", ContinueOnError.Value ? "true" : "false");
+        if (TimeoutMinutes.HasValue)
+            Scalar("timeout-minutes", TimeoutMinutes.Value.ToString(CultureInfo.InvariantCulture));
+
+        return;
+
+        void Scalar(string key, string value)
+        {
+            writer.WriteLine((written ? "  " : "- ") + $"{key}: {value}");
```
ChrisonSimtian, line 71: **Values are emitted unquoted, unlike the built-in steps.** `Scalar` renders `{key}: {value}` raw, but `GitHubActionsCacheStep`/`GitHubActionsRunStep` deliberately quote names — e.g. `"- name: " + $"Cache: …".SingleQuote()` — precisely because the value contains a colon. As written, `Name = "Deploy: prod"` renders `name: Deploy: prod` (invalid YAML), and the same applies to any `If`/`Uses` value with a colon-space, `#`, or a leading `@`/`*`/quote.

The generation-time validation covers structure (uses XOR run, `with`⇒`uses`, `shell`⇒run) but not value safety, so the "never emits invalid YAML" guarantee doesn't hold for free-text scalars. Suggest routing `name` (and ideally other free-text scalars) through the same quote helper the subsystem already uses, plus a spec with a colon in `Name` to pin it. `run:`/`env:` staying raw is fine — that matches the built-in run step.

src/Fallout.Common/CI/GitHubActions/Configuration/GitHubActionsCustomStep.cs
```diff
@@ -0,0 +1,80 @@
...
+/// <see cref="Uses"/> renders a marketplace/action step; a non-empty <see cref="Run"/> renders a shell
+/// step (a single entry as <c>run: x</c>, multiple as a <c>run: |</c> block scalar). Exactly one of the
+/// two must be set — enforced at generation time.
+/// </summary>
+public class GitHubActionsCustomStep : GitHubActionsStep
+{
+    public string Name { get; set; }
+    public string Uses { get; set; }
+    public Dictionary<string, string> With { get; set; } = new Dictionary<string, string>();
+    public Dictionary<string, string> Env { get; set; } = new Dictionary<string, string>();
```
ChrisonSimtian, line 21: `Dictionary<string,string>` enumeration order isn't guaranteed, so multi-entry `with:`/`env:` blocks can render in varying order → non-reproducible generated YAML and potential snapshot churn. The current specs only exercise single-entry dicts, so it's latent. Consider preserving insertion order (or sorting keys) for stable output.

src/Fallout.Common/CI/GitHubActions/Configuration/GitHubActionsStepPipeline.cs
```diff
@@ -0,0 +1,51 @@
...
+    public string WorkflowName { get; }
+
+    /// <summary>The runner image of this job.</summary>
+    public GitHubActionsImage Image { get; }
+
+    /// <summary>A read-only view of the built-in steps already assembled for this job.</summary>
+    public IReadOnlyList<GitHubActionsStep> BuiltInSteps { get; }
+
+    /// <summary>Insert one custom step at <paramref name="position"/>. Multiple inserts at one position render in call order.</summary>
+    public void Insert(GitHubActionsStepPosition position, GitHubActionsCustomStep step)
```
ChrisonSimtian, line 34: Minor: a `null` step isn't guarded here, so it surfaces later as an `NullReferenceException` in `ValidateCustomSteps` (`step.Name`) rather than the clean `ArgumentException` the design otherwise promises. A cheap `Assert.NotNull(step)` here would keep the failure mode consistent.

ChrisonSimtian (approved):
Really clean, additive design — the generator keeps sole control of the base sequence and the hook only inserts. Test coverage is the standout: all four positions, the cache/artifacts-absent anchoring case, ordering, workflow scoping, and a byte-identical no-op regression guard.

Approving — nothing here blocks, but one is worth fixing before this gets real-world use (inline): custom-step scalar values are emitted unquoted, whereas every built-in step quotes names, so a `Name`/`If`/`Uses` containing a colon-space, `#`, or leading indicator char emits invalid YAML. That contradicts the "never emits invalid YAML" line, which today only covers the structural uses/run checks. Two smaller notes on dictionary ordering and a null-step guard.

Confirmed `Assert.True` throws `ArgumentException` (`src/Fallout.Utilities/Assert.cs`), so the validation specs assert the right type. 👍

tests/Fallout.Common.Specs/CI/GitHubActionsCustomStepSpecs.cs
avidenic: The raw literal keeps the source file's line endings (C# preserves them as-written), but the assert compares against the rendered output, and that's written with Environment.NewLine, so CRLF on Windows, regardless of the literal. On the CI runners the two line up because each runner checks out with its OS default. They will not line up with `autocrlf=input` (an LF checkout on Windows) they will render CRLF vs an LF in `""" """`. Normalizing both sides to `\n` makes the assert independent of git line-ending config. Happy to drop it and rely on runner defaults if you'd rather keep the test lean.

src/Fallout.Common/CI/GitHubActions/Configuration/GitHubActionsCustomStep.cs
avidenic: Fixed in cf18d3f7.

src/Fallout.Common/CI/GitHubActions/Configuration/GitHubActionsCustomStep.cs
avidenic: Fixed in cf18d3f7.

src/Fallout.Common/CI/GitHubActions/Configuration/GitHubActionsStepPipeline.cs
avidenic: Fixed in cf18d3f7.

src/Fallout.Common/CI/GitHubActions/GitHubActionsAttribute.cs
```diff
@@ -235,61 +235,98 @@ protected virtual GitHubActionsJob GetJobs(GitHubActionsImage image, IReadOnlyCo
...
+                                  Path = Build.RootDirectory.GetUnixRelativePathTo(artifact),
+                                  Condition = PublishCondition
+                              });
+        }
+
+        var builtInSteps = new List<GitHubActionsStep> { checkout };
+        if (cache != null)
+            builtInSteps.Add(cache);
+        builtInSteps.Add(run);
+        builtInSteps.AddRange(artifacts);
```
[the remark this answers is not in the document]
avidenic: `BuiltInSteps` deliberately exposes the concrete step objects so a `ConfigureSteps` implementation can pattern-match a job's shape — that's the #336 (job-DAG) forward-compat use, and it needs the real types, not opaque handles. There's no consumer in this PR, so rather than deep-clone every step per job now, the contract is inspect-only; defensive copies can land when #336 actually reads it.

src/Fallout.Common/CI/GitHubActions/Configuration/GitHubActionsStepPipeline.cs
```diff
@@ -0,0 +1,51 @@
...
+            _inserts[position] = list = new List<GitHubActionsCustomStep>();
+        list.Add(step);
+    }
+
+    /// <summary>Insert several custom steps at <paramref name="position"/>, in enumeration order.</summary>
+    public void Insert(GitHubActionsStepPosition position, IEnumerable<GitHubActionsCustomStep> steps)
+    {
+        foreach (var step in steps)
+            Insert(position, step);
+    }
```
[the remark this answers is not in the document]
avidenic: Fixed in cf18d3f7 — both `Insert` overloads now `Assert.NotNull(...)` before storing. It throws `ArgumentException` (via the subsystem's `Assert` helper) rather than `ArgumentNullException`, to stay consistent with the rest of the generator's validation, which the specs already assert on. Added a null-step spec.

src/Fallout.Common/CI/GitHubActions/Configuration/GitHubActionsStepPipeline.cs
```diff
@@ -0,0 +1,51 @@
...
+
+    internal IReadOnlyList<GitHubActionsCustomStep> GetInserts(GitHubActionsStepPosition position)
+        => _inserts.TryGetValue(position, out var list) ? list : (IReadOnlyList<GitHubActionsCustomStep>)new GitHubActionsCustomStep[0];
```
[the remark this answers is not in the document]
avidenic: Left as `new T[0]` to match the subsystem convention (every empty-array initializer in CI/ uses `new T[0]`, not `Array.Empty<T>()`). GetInserts is called four times per job with the empty branch only hit for positions that got no inserts, so the allocation is negligible.

src/Fallout.Common/CI/GitHubActions/GitHubActionsAttribute.cs
```diff
@@ -235,61 +235,98 @@ protected virtual GitHubActionsJob GetJobs(GitHubActionsImage image, IReadOnlyCo
...
+        foreach (var step in pipeline.AllInserts)
+        {
+            var id = step.Name ?? step.Uses ?? "(run step)";
+            var hasUses = !step.Uses.IsNullOrWhiteSpace();
+            var hasRun = step.Run.Any(x => !x.IsNullOrWhiteSpace());
+
+            Assert.True(hasUses ^ hasRun,
+                $"Custom step '{id}' in workflow '{_name}' must set exactly one of '{nameof(GitHubActionsCustomStep.Uses)}' or '{nameof(GitHubActionsCustomStep.Run)}'");
+            Assert.True(step.With.Count == 0 || hasUses,
+                $"Custom step '{id}' in workflow '{_name}' sets '{nameof(GitHubActionsCustomStep.With)}' but no '{nameof(GitHubActionsCustomStep.Uses)}'; 'with:' is only valid on a 'uses:' step");
```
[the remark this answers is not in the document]
avidenic: Fixed in cf18d3f7. Validation now reads `Run` via `is { } run` and `With` via `(With?.Count ?? 0)`, so null collections are treated as empty and produce the normal `ArgumentException` path (e.g. null `Run` + null `Uses` → the clean "exactly one of Uses/Run" error) instead of an NRE. Added a null-collections spec.

src/Fallout.Common/CI/GitHubActions/Configuration/GitHubActionsCustomStep.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            Scalar("run", Run[0]);
+        }
+        else if (Run.Length > 1)
+        {
+            writer.WriteLine((written ? "  " : "- ") + "run: |");
+            written = true;
+            using (writer.Indent())
+                Run.ForEach(x => writer.WriteLine($"  {x}"));
+        }
```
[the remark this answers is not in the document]
avidenic: Fixed in cf18d3f7. `Write` guards `With`/`Env` with `(x?.Count ?? 0) > 0` and `Run` with `Run?.Length ?? 0`, so null collections render as empty. And `name` is now single-quoted like the built-in steps, so a `:` in the name stays valid YAML. Specs added for both.

src/Fallout.Common/CI/GitHubActions/Configuration/GitHubActionsCustomStep.cs
avidenic: Follow-up in 5b819594: moved name quoting off `SingleQuote()` onto a new `SingleQuoteYaml()` in `String.Quoting.cs`, which escapes an embedded quote by doubling it (`''`) per the YAML spec. `SingleQuote()`'s backslash form (`\'`) is invalid inside a YAML single-quoted scalar, so `Name = "Bob's step"` would have emitted `'Bob\'s step'`. Output is identical for names without a quote, so no snapshot churn.

Correcting my earlier note here: it no longer routes through `SingleQuote()`. The built-in step names still do and carry the same latent issue (safe today only because those names are controlled), so I left the shared helper untouched (26 call sites across shell/log/YAML) and filed #483 to migrate the CI writers to `SingleQuoteYaml()` separately.

tests/Fallout.Common.Specs/CI/GitHubActionsCustomStepSpecs.cs
ITaluone: Another way you could handle this: use Verify (the most tests for any sort of generator using Verify already, would fit well into existing testing strategy)

Save the .verified file with LF and/or add this line to the .gitattributes: 
`*.verified.* text eol=lf`

For me the verify approach would be the way to go, because you do not have to manually set up the expected multiline string, just give the received file a look and accept if cool

src/Fallout.Utilities/Text/String.Quoting.cs
```diff
@@ -71,6 +71,16 @@ public static string SingleQuote(this string str)
...
 
+    /// <summary>
+    /// Single-quotes a given string as a YAML flow scalar, escaping an embedded single quote by doubling it
+    /// (<c>''</c>) per the YAML spec. Prefer this over <see cref="SingleQuote"/> when emitting YAML: the latter
+    /// backslash-escapes, which is fine for shell/log output but invalid inside a YAML single-quoted scalar.
+    /// </summary>
+    public static string SingleQuoteYaml(this string str)
+    {
+        return $"'{str?.Replace("'", "''")}'";
+    }
```
ITaluone, line 82: What about the case where a user already has thought about "Hey, this could be invalid yaml, fix it prematurely"?

The he would end up e.g. with `'Already qouted'` -> `'''Already quoted'''`  (which is technically valid yaml, but a bit ugly

Maybe do this:

```suggestion
    public static string SingleQuoteYaml(this string str)
    {
        return $"'{str?.Trim('\').Replace("'", "''")}'";
    }
```

ITaluone (approved):
Nice work!

Just this two questions

src/Fallout.Utilities/Text/String.Quoting.cs
avidenic: Not a bad idea, but since this is new API it might be premature optimisation IMHO. We might also be removing characters user would legitimately want in YAML.

tests/Fallout.Common.Specs/CI/GitHubActionsCustomStepSpecs.cs
avidenic: Excellent suggestion. Done in 3acf6fde - switched GitHubActionsCustomStepSpecs to Verify (one snapshot per case), matching the generator strategy used elsewhere. Verify normalizes line endings on compare, so the `ShouldRenderAs` helper and the whole CRLF question are gone. Also added `*.verified.* text eol=lf` to .gitattributes to keep the snapshots LF in the repo.

ITaluone: Premium :)

thanks

src/Fallout.Utilities/Text/String.Quoting.cs
ITaluone: Ok.. nevermind... just curious if this is a real problem (or "just" a problem in my head) 😂

ITaluone (approved):
Very nice work!

Thanks again

PR #448, opened 2026-06-30
Introduce IFalloutCommand dispatch with DI for Fallout.Cli (#392, PR 0)

ChrisonSimtian:
Replaces the reflection-over-`Program` CLI dispatch with one-command-per-type behind a typed `IFalloutCommand`, resolved via Microsoft.Extensions.DependencyInjection. PR 0 of an incremental series — lands the abstraction, dispatcher, and first conversion; a transitional adapter keeps the other 13 handlers working so nothing regresses.

### What changed
- `IFalloutCommand` + `CommandDispatcher` replace `Program.Handle`; resolution stays dash/case-insensitive (every spelling the old reflection accepted).
- `IConsolePrompts` / `SpectreConsolePrompts` — Spectre prompt helpers lifted off `Program` into an injectable service.
- `RunCommand` — first real conversion. `DelegateCommand` adapts the 13 still-legacy handlers so dispatch is uniform; each follow-up PR deletes one.
- Reflection dispatch removed; `CommandDispatcherTests` add first dispatch coverage.

### Why it's not breaking
`Fallout.Cli` is the tool Exe, not a consumed library; the `:command` surface and shell-function names are preserved exactly. Hence `target/2026` / `main`, no `breaking-change` label.

### Verification
`dotnet build src/Fallout.Cli` clean · `Fallout.Cli.Tests` 30 pass · `fallout :bogus` prints the full command manifest.

Follow-ups (#392): setup · update · add-package · cake-convert/clean · complete · get-configuration · secrets · trigger · navigation → then collapse `Program` to a thin entry point. Raised sequentially off this branch as it lands.

Part of #392.

---
_Re-raised from the fork (`ChrisonSimtian:cli-command-dispatch-foundation`) to move the branch off the main repo. Supersedes #394; same history and diff. Bottom of the cli-cmd stack — follow-ups open one at a time as each merges._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #392, opened 2026-06-17
#Refactor CLI command dispatch: replace reflection-based partial `Program` god-class with `IFalloutCommand` types


ChrisonSimtian:
### Problem
`Fallout.Cli` implements every CLI command as a `static int` method on one `partial class Program`, split across ~10 `Program.*.cs` files and dispatched by reflection (`Program.Handle` reflects over its own `int`-returning methods). The partial split looks organized, but the underlying type is a god object — it owns dispatch, secrets, packages, Cake interop, navigation, config, run orchestration, and the Spectre prompts. Reflection dispatch is stringly-typed: a typo'd signature compiles and fails at runtime, and nothing is DI-able or unit-testable in isolation.

### Outcome
One command = one type behind a typed `IFalloutCommand`, discovered by type via DI instead of reflection. The `:command` CLI surface and shell-function names stay exactly as they are (non-breaking).

```csharp
public interface IFalloutCommand
{
    string Name { get; }   // e.g. "run", "setup"
    int Execute(string[] args, AbsolutePath rootDirectory, AbsolutePath buildScript);
}
```

Each `Program.X.cs` partial becomes a standalone `XCommand` implementing `IFalloutCommand`; the dispatcher resolves by `Name`. The `partial` keyword disappears. Shared helpers move into injectable services in the final collapse PR.

### Acceptance criteria
- [ ] `IFalloutCommand` + a type/DI-based dispatcher replace reflection dispatch
- [ ] Every command is its own type; `Program` collapses to a thin entry point (no `partial`)
- [ ] CLI command names + shell-function spellings unchanged (dash/case-insensitive preserved)
- [ ] Dispatch has unit-test coverage
- [ ] Shared helpers extracted into injectable services

### Approach
Incremental — one command per PR behind a transitional `DelegateCommand` adapter so nothing regresses. Tracked by the #394 stack.

ChrisonSimtian:
Claude picked wolverine over the license, Im tempted to just pick MediatR. Why? I use it at work and Im familiar with it. Just need to make sure the license terms are a match for this project

avidenic:
> Claude picked wolverine over the license, Im tempted to just pick MediatR. Why? I use it at work and Im familiar with it. Just need to make sure the license terms are a match for this project

This one is almost a drop in replacement for MediatR -> https://github.com/martinothamar/Mediator

ChrisonSimtian:
[...]

Issue #26, opened 2026-05-18
[P1] Update README, CLAUDE.md, CONTRIBUTING.md, CODE_OF_CONDUCT.md to Fallout brand


ChrisonSimtian:
## Scope

Cosmetic — text and link updates only, no code changes.

### Files
- \`README.md\` — title, tagline, badges, build-status table, references, Twitter/Slack links
- \`CLAUDE.md\` — \"what this project is\" header, links
- \`CONTRIBUTING.md\` — replace \"NUKE is a personal project that was made open-source...\" with new ownership story
- \`CODE_OF_CONDUCT.md\` — references to NUKE

### Preserve attribution

In README, add a "Based on NUKE" section credit:
> Fallout is the successor to NUKE, originally created by Matthias Koch and many contributors. Fallout continues NUKE's mission as a C#-first build automation framework for .NET.

### Badges

NuGet badges need to come down until we have published \`Fallout.*\` packages on nuget.org. Replace with GitHub Packages badges or remove for now.

## Done when

- [ ] All four files updated
- [ ] Original NUKE attribution prominent in README
- [ ] No badges link to broken/wrong URLs

Issue #17, opened 2026-05-18
Azure DevOps: incorrect GitRepository branch value with multi-repo pipelines — mirrors upstream #1587


ChrisonSimtian:
Mirror of [nuke-build/nuke#1587](https://github.com/nuke-build/nuke/issues/1587). The Azure Pipelines GENERATOR (still ships in Nuke.Common/CI for users) reports the wrong branch when the pipeline uses multiple repositories.

PR #476, opened 2026-07-11
Convert remaining Fallout.Cli commands to IFalloutCommand dispatch (#392)

ChrisonSimtian:
Second step of the `Fallout.Cli` command-dispatch migration ([#392](https://github.com/Fallout-build/Fallout/issues/392)). Builds on the merged foundation (#448); **supersedes fork-only PRs #17–#26**, which were collapsed into one and re-authored against the current (async/internal) contract.

Converts the remaining legacy `Program.X` handlers into real `IFalloutCommand` types and reduces `Program` to a thin entry point.

## What
- `complete`, `get-configuration`, `add-package`, `update`, `secrets`, `cake-convert`, `cake-clean`, `setup`, and the `navigation` cluster → real command types (`ExecuteAsync`, DI-injected services)
- New injectable services: `ConfigurationReader`, `PackageManager`, `BuildScaffolder`, `CakeConverter`, `CliConventions`
- Deletes the `DelegateCommand` shims + every `Program.X.cs` partial
- Specs re-homed to `Fallout.Cli.Specs` (async); existing specs repointed to the new service homes

## Not breaking / no regressions
Derived from `main`'s current handlers, so these are preserved: `.slnx` setup support, `net10.0` target, the `_FALLOUT_VERSION_` template token, and the nuke→fallout rebrand. `Fallout.Cli.Specs` 46/46 green; runtime dispatch verified (all 15 commands resolve via DI).

## Notes
- Not per-command commits: the shared `Program` registration + config services couple the commands (`CakeConvert` depends on `SetupCommand`), so it lands as 4 logical commits.
- Sync commands wrap in `Task.FromResult(...)` (as the existing `DelegateCommand` did); only genuinely async work is truly async.


tests/Fallout.Cli.Specs/Commands/CakeCleanCommandSpecs.cs
```diff
@@ -0,0 +1,22 @@
...
+using FluentAssertions;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Fallout.Cli.Specs.Commands;
+
+public class CakeCleanCommandSpecs
+{
+    [Fact]
+    public void Name_IsCakeClean()
```
ITaluone: Fix test names according to convention

src/Fallout.Cli/Commands/AddPackageCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+    public AddPackageCommand(IConfigurationReader configuration, IPackageManager packages)
+    {
+        _configuration = configuration;
+        _packages = packages;
+    }
+
+    public string Name => "add-package";
+
+    public Task<int> ExecuteAsync(string[] args, AbsolutePath rootDirectory, AbsolutePath buildScript)
+        => Task.FromResult(Execute(args, rootDirectory, buildScript));
```
ITaluone: When doing `async` here you can make the private `ExecuteAsync` async so you in return can remove the sync-over-async anti-pattern in line 42.

Applies also to other commands...

ITaluone (changes requested):

tests/Fallout.Cli.Specs/Commands/CakeCleanCommandSpecs.cs
ChrisonSimtian: Done — renamed all six command specs to the AV1600 behavior-focused style (present-tense observable behavior, e.g. `Command_is_named_cake_clean`, `Declining_the_deletion_prompt_deletes_nothing_and_returns_zero`). 46/46 still green.

src/Fallout.Cli/Commands/AddPackageCommand.cs
ChrisonSimtian: Fixed. Folded the private `Execute` into an `async ExecuteAsync` and `await` the work directly, dropping `.GetAwaiter().GetResult()` in `AddPackageCommand`, `CakeConvertCommand`, and `SetupCommand` (the three commands that actually do async work).

The purely-sync commands (Complete, GetConfiguration, Secrets, Trigger, Update, CakeClean, Navigation) keep `Task.FromResult(Execute(...))` — they have nothing to await, so forcing `async` would only add a CS1998 no-await warning.

One deliberate exception: `ProjectUpdater.Update` (used by `UpdateCommand`) still blocks on `GetLatestVersion`. It is a `public static` helper consumed through the synchronous `IConsolePrompts.ConfirmExecution(Action)` contract, so making it async would ripple an async overload through the prompt infra — out of scope for this PR. Happy to do it as a follow-up if you want it.

ITaluone (approved):

ITaluone (approved):
Nice!

src/Fallout.Cli/Commands/UpdateCommand.cs
```diff
@@ -63,8 +65,65 @@ private async Task UpdateBuildProjectAsync(AbsolutePath buildScript)
...
+
+        if (previousPackageVersion.MinVersion >= NuGetVersion.Parse("0.23.5"))
+            RemoveLegacyFileIncludes(buildProject);
+
+        buildProject.Save();
+    }
+
+    private static void UpdateTargetFramework(Microsoft.Build.Evaluation.Project buildProject)
+    {
+        buildProject.SetProperty("TargetFramework", "net8.0");
```
ITaluone: I think this should be `net10.0`?

I recently added Fallout (via 10.3.49 global tool). It had the `<TargetFramework>net8.0</TargetFramework>` and it wasn't able to restore the packages and/or find all required namespaces (like Fallout.Common).

ITaluone (changes requested):

ITaluone:
Found one little quirck :D

src/Fallout.Cli/Commands/UpdateCommand.cs
ITaluone: I tried an explicit project with net8.0.. even this isn't working.. so this have to be (under all circumstances) net10.0..

ChrisonSimtian: yes and probably also a more central config param, not hidden away somewhere deep down 😅

ITaluone: idk... is this needed at all? Given, that a different version than net10.0 won't work anyway?

Or do you mean for us maintainers, when a new target framework comes out?

ChrisonSimtian: the latter. Imagine dotnet12 and you have to go to 30 places to change this. Might as well do 1.but lets do a dedicated cleanup PR for that at some point :-)

ChrisonSimtian: Fixed in 602091e6 — `update` now writes `net10.0` (matching `SetupCommand`'s `TARGET_FRAMEWORK`). Centralizing the shared literal into one config param is tracked for the dedicated cleanup PR as agreed.

PR #478, opened 2026-07-11
Recognizes version variables and updates the version accordingly

### LINKED ISSUE #477: Fallout.Migrate: Also recognize csproj variables
### Description

When doing fallout-migrate on a project which does have mulitple Fallout.* packages installed, and using a shared variable for the version, like:

```xml

<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net10.0</TargetFramework>
    <RootNamespace></RootNamespace>
    <NoWarn>CS0649;CS0169</NoWarn>
    <FalloutRootDirectory>..\</FalloutRootDirectory>
    <FalloutScriptDirectory>..\</FalloutScriptDirectory>
    <NukeVersion>10.3.49</NukeVersion >
    <FalloutTelemetryVersion>1</FalloutTelemetryVersion>
  </PropertyGroup>
  <ItemGroup>
    <PackageDownload Include="GitVersion.Tool" Version="[6.5.1]" />
    <PackageDownload Include="ReportGenerator" Version="[5.5.0]" />
    <PackageDownload Include="xunit.runner.console" Version="[2.9.2]" />
    <PackageReference Include="LibGit2Sharp" Version="0.31.0" />
    <PackageReference Include="Nuke.Common" Version="$(NukeVersion)" />
    <PackageReference Include="Nuke.Components" Version="$(NukeVersion)" />
    <PackageReference Include="SharpCompress" Version="0.48.0" />
    <!-- Replace the implicitly referenced version 6.12.1, which is defined by the package Fallout.Common and has known vulnerabilities, with a fixed version -->
    <PackageReference Include="Nuget.Packaging" Version="7.3.1" />
  </ItemGroup>
</Project>

```

Neither the variable "NukeVersion" is converted, nor the version is replaced. 

### Usage Example

Just doing `fallout-migrate` and have this automagically done

### Alternative

Doing it yourself

Yes

IT-VBFK:
## Summary

Update `Fallout.Migrate` so that package versions referenced through MSBuild properties are migrated to the target Fallout version.

Previously, the `.csproj` rewriter handled package references with literal versions:

```xml
<PackageReference Include="Nuke.Common" Version="10.3.49" />
```

However, projects that shared a version through an MSBuild property were only partially migrated:

```xml
<PropertyGroup>
  <NukeVersion>10.3.49</NukeVersion>
</PropertyGroup>

<ItemGroup>
  <PackageReference Include="Nuke.Common" Version="$(NukeVersion)" />
  <PackageReference Include="Nuke.Components" Version="$(NukeVersion)" />
</ItemGroup>
```

The package IDs could be renamed to `Fallout.*`, while the property and its value remained on the old NUKE version.

This left the migrated project in an inconsistent state and could cause package restore or build failures.

Closes #477.

### Distinguish literal versions from property-based versions

The inline package-version rewrite now excludes version values beginning with an MSBuild property expression such as:

```xml
Version="$(NukeVersion)"
```

This prevents the existing literal-version rewrite from replacing the property reference itself with an inline version.

Property-based version management therefore remains property-based after migration.

For example:

```xml
<PackageReference Include="Nuke.Common" Version="$(NukeVersion)" />
```

is processed by the property-aware migration flow instead of being flattened into:

```xml
<PackageReference Include="Fallout.Common" Version="11.0.0" />
```

### Detect version properties used by package references

A new rewrite pass discovers MSBuild properties referenced by `PackageReference` version attributes:

```xml
<PackageReference Include="Nuke.Common" Version="$(PropertyName)" />
```

The referenced property names are deduplicated and used to locate the corresponding property elements in the project file.

### Update referenced property values

For each discovered version property, the rewriter updates its value to the requested Fallout version.

A NUKE-specific version property is migrated completely.

Before:

```xml
<PropertyGroup>
  <NukeVersion>10.3.49</NukeVersion>
</PropertyGroup>

<ItemGroup>
  <PackageReference Include="Nuke.Common" Version="$(NukeVersion)" />
  <PackageReference Include="Nuke.Components" Version="$(NukeVersion)" />
</ItemGroup>
```

After:

```xml
<PropertyGroup>
  <FalloutVersion>11.0.0</FalloutVersion>
</PropertyGroup>

<ItemGroup>
  <PackageReference Include="Fallout.Common" Version="$(FalloutVersion)" />
  <PackageReference Include="Fallout.Components" Version="$(FalloutVersion)" />
</ItemGroup>
```

The existing MSBuild-property migration renames `NukeVersion` to `FalloutVersion`, while the new rewrite updates the property's value.

### Preserve custom property names

Projects are not required to name their shared version property `NukeVersion`.

When a custom property is used, its name is preserved and only its value is updated.

Before:

```xml
<PropertyGroup>
  <CiProjectVersion>10.3.49</CiProjectVersion>
</PropertyGroup>

<ItemGroup>
  <PackageReference Include="Nuke.Common" Version="$(CiProjectVersion)" />
</ItemGroup>
```

After:

```xml
<PropertyGroup>
  <CiProjectVersion>11.0.0</CiProjectVersion>
</PropertyGroup>

<ItemGroup>
  <PackageReference Include="Fallout.Common" Version="$(CiProjectVersion)" />
</ItemGroup>
```

This keeps the project's existing version-management structure intact while ensuring that migrated Fallout packages use the correct version.

### Leave unrelated properties untouched

Only properties referenced by a `PackageReference` version attribute are selected for updating.

Other version-like properties in the same project are left unchanged, avoiding broad replacements based solely on property names or values.

## Migration behavior

| Input | Result |
|---|---|
| Literal package version | Package ID and inline version are updated |
| Package version using `$(NukeVersion)` | Property and references are renamed to `FalloutVersion`, and the property value is updated |
| Package version using a custom property | The property name is preserved and its value is updated |
| Unreferenced property | Left unchanged |
| Unrelated package reference | Left unchanged |

## Tests

Added regression coverage for the scenarios reported in #477:

- A shared `NukeVersion` property is renamed to `FalloutVersion`.
- All package references using `$(NukeVersion)` are updated to use `$(FalloutVersion)`.
- The renamed property receives the requested Fallout version.
- An arbitrary shared version property keeps its original name while its value is updated.
- An unrelated, unreferenced property remains unchanged.
- Multiple NUKE package references sharing the same property are migrated consistently.

## Result

Running `fallout-migrate` now produces a consistent `.csproj` for projects that manage NUKE package versions through MSBuild properties.

Both the package references and the property supplying their version are migrated automatically, without requiring users to manually repair the project afterward.

🤖 helped out here :D


src/Fallout.Migrate/CsprojRewriter.cs
```diff
@@ -20,13 +23,18 @@ internal static class CsprojRewriter
     private static readonly Regex PackageReferencePattern =
         new(@"(?<=\b(?:Include|Update|Remove)="")Nuke\.(?=[A-Z])", RegexOptions.Compiled);
 
+    // Detects MSBuild variables used in PackageReference Version attributes: Version="$(MyVar)"
+    private static readonly Regex PackageReferenceVariablePattern = new(
+        @"<PackageReference\s+[^>]*?Version=""\$\((?<variable>[^)]+)\)""",
+        RegexOptions.Compiled);
```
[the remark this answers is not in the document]
ChrisonSimtian: @IT-VBFK this one might be quite critical.
TLDR; your regex might grab non-nuke references and migrate them. Can you a) cover that with a test please and b) fix if copilot is right?

ChrisonSimtian (changes requested):

src/Fallout.Migrate/CsprojRewriter.cs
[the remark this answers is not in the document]
ITaluone: Thanks for hitting this. This was a rel problem --> fixed now

ITaluone:
@copilot review

src/Fallout.Migrate/CsprojRewriter.cs
```diff
@@ -68,27 +69,28 @@ public static RewriteResult Rewrite(string original, string falloutVersion)
...
-        content = PackageReferencePattern.Replace(content, _ =>
+        content = packageReferencePattern.Replace(content, _ =>
         {
             edits++;
             return "Fallout.";
         });
 
-        content = MSBuildPropertyPattern.Replace(content, _ =>
+        content = msBuildPropertyPattern.Replace(content, _ =>
+
```
ITaluone:
```suggestion
```

ChrisonSimtian (approved):

ITaluone:
Do not merge, please (meaning now).

I rebase the commits later to better commits :)

src/Fallout.Migrate/CsprojRewriter.cs
```diff
@@ -79,6 +88,33 @@ public static RewriteResult Rewrite(string original, string falloutVersion)
...
+        {
+            var escapedVariable = Regex.Escape(variable);
+            var pattern = $@"(?<open><{escapedVariable}(?=\s|>)[^>]*>)(?<value>[^<]+)(?<close></{escapedVariable}\s*>)";
+            content = Regex.Replace(content, pattern, m =>
+            {
+                var currentValue = m.Groups["value"].Value;
+                if (currentValue != falloutVersion)
+                {
+                    edits++;
+                    return m.Groups["open"].Value + falloutVersion + m.Groups["close"].Value;
```
ITaluone:
```suggestion
                    return $"{m.Groups["open"].Value}{falloutVersion}{m.Groups["close"].Value}";
```

tests/Fallout.Migrate.Specs/CsprojRewriterSpecs.cs
```diff
@@ -182,4 +182,94 @@ public void LeavesOtherSystemPackagesAlone()
...
+                                 <PackageReference Include="Nuke.Components" Version="$(NukeVersion)" />
+                               </ItemGroup>
+                             </Project>
+                             """;
+
+        var result = CsprojRewriter.Rewrite(input, TestFalloutVersion);
+
+        result.Content.Should().NotContain("NukeVersion")
+            .And.Contain("$(FalloutVersion)", Exactly.Twice())
+            .And.Contain("<FalloutVersion>11.0.0</FalloutVersion>");
```
ITaluone: To Verify...

ITaluone (changes requested):

ITaluone (approved):

ITaluone:
@avidenic Can you do the final review.. mine isn't worth it (person from IT-VBFK == ITaluone) ;)

dennisdoomen:
As I'm cleaning up the migration, please wait with the merge after which I've taken a look.

src/Fallout.Migrate/CsprojRewriter.cs
```diff
@@ -79,6 +87,32 @@ public static RewriteResult Rewrite(string original, string falloutVersion)
             return string.Empty;
         });
 
+        // Pass 4 — extract variables used in PackageReferences and bump their properties.
+        var variablesToBump = packageReferenceVariablePattern.Matches(content)
+            .Select(m => m.Groups["variable"].Value)
+            .ToHashSet();
+
+        // Always include NukeVersion/FalloutVersion as they are canonical.
+        variablesToBump.Add("NukeVersion");
```
avidenic: I think we already changed nuke to fallout in pass 2 so no remaining nuke versions should be here. So the line `variablesToBump.Add("NukeVersion");` can probably be dropped.

src/Fallout.Migrate/CsprojRewriter.cs
```diff
@@ -79,6 +87,32 @@ public static RewriteResult Rewrite(string original, string falloutVersion)
...
+            content = Regex.Replace(content, pattern, m =>
+            {
+                var currentValue = m.Groups["value"].Value;
+                if (currentValue != falloutVersion)
+                {
+                    edits++;
+                    return $"{m.Groups["open"].Value}{falloutVersion}{m.Groups["close"].Value}";
+                }
+
+                return m.Value;
```
avidenic: One thing to note - we renamed nuke to fallout in pass 2 then extracted the variable for `Fallout.*` in line 91. It is highly unlikely that someone would use the same variable for non nuke packages, but it can be possible. So we should either note this somewhere or add some kind of guard for this. To illustrate: 

```xml
<PropertyGroup>
  <PkgVersion>10.1.0</PkgVersion>
</PropertyGroup>
<ItemGroup>
  <PackageReference Include="Nuke.Common" Version="$(PkgVersion)" />
  <PackageReference Include="Some.ThirdParty" Version="$(PkgVersion)" />
</ItemGroup>
```

`Some.ThirdParty` would then possibly get non existing version.

src/Fallout.Migrate/CsprojRewriter.cs
```diff
@@ -20,13 +23,18 @@ internal static class CsprojRewriter
     private static readonly Regex packageReferencePattern =
         new(@"(?<=\b(?:Include|Update|Remove)="")Nuke\.(?=[A-Z])", RegexOptions.Compiled);
 
+    // Detects MSBuild variables used for Fallout.* package versions: Version="$(MyVar)"
+    private static readonly Regex packageReferenceVariablePattern = new(
+        @"<PackageReference\s+[^>]*?\b(?:Include|Update)=""Fallout\.[^""]+""[^>]*?\bVersion=""\$\((?<variable>[^)]+)\)""",
+        RegexOptions.Compiled);
```
avidenic: Parsing xmls with regex has one downside - since we're working with text and not xml nodes, we miss things like:

- version before include: `<PackageReference Version="$(MyVer)" Include="Fallout.Common" />` 
- composite versions `Version="$(Major).$(Minor)"`

and probably some others as well. Might be worth another issue. Not sure this PR should fix edge cases like this - or we should at least note shortcomings somewhere.

avidenic (approved):
Good improvements, some small issues to look at. Some should probably go into follow up issues / PRs. Or we can decide we're fine and just add docs about gaps.

src/Fallout.Migrate/CsprojRewriter.cs
ITaluone: Maybe this whole CsprojRewrite thing needs a rewrite to full xml editing instead of plain text edit.

For example: 80cc687

src/Fallout.Migrate/CsprojRewriter.cs
ITaluone: True.

What about decouple things in such a case? Like so:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <FalloutVersion>11.0.0</FalloutVersion>
    <PkgVersion>10.1.0</PkgVersion>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Fallout.Common" Version="$(FalloutVersion)" />
    <PackageReference Include="Some.ThirdParty" Version="$(PkgVersion)" />
  </ItemGroup>
</Project>
```

See 5d3adcb for example

src/Fallout.Migrate/CsprojRewriter.cs
avidenic: That is a substantial rewrite, let me check the full diff now.

ITaluone: yeah I know :) at least tests stays green :D

src/Fallout.Migrate/CsprojRewriter.cs
```diff
@@ -1,84 +1,286 @@
...
+    private const string VersionAttribute = "Version";
+    private const string FalloutVersionProperty = "FalloutVersion";
+    private const string PropertyGroupElement = "PropertyGroup";
 
     public static RewriteResult Rewrite(string original, string falloutVersion)
+    {
+        XDocument doc;
+        try
+        {
+            doc = XDocument.Parse(original, LoadOptions.None);
```
avidenic: Just a small nit, if we ever touch a file and the whitespace is different than in our test, e.g. 4 it will produce a big diff instead of surgical changes. If possible, I would use `LoadOptions.PreserveWhitespace`

src/Fallout.Migrate/CsprojRewriter.cs
```diff
@@ -1,84 +1,286 @@
...
+        {
+            FalloutVersionProperty
+        };
+
+        edits += DecoupleVariables(elements, variablesToBump);
+        edits += InjectFalloutVersion(elements, falloutVersion);
+        edits += BumpVariableValues(elements, variablesToBump, falloutVersion);
+
+        var content = (doc.Declaration != null ? doc.Declaration + Environment.NewLine : string.Empty) +
+                      doc.ToString(SaveOptions.None);
```
avidenic: To expand my previous comment and close the whitespace issue, I would use `SaveOptions.DisableFormatting` here.

src/Fallout.Migrate/CsprojRewriter.cs
```diff
@@ -1,84 +1,286 @@
...
-        "(?:RootDirectory|ScriptDirectory|TelemetryVersion|BaseDirectory|BaseNamespace|" +
-        "UseNestedNamespaces|RepositoryUrl|UpdateReferences|ContinueOnError|TaskTimeout|" +
-        "Timeout|TasksEnabled|DefaultExcludes|ExcludeBoot|ExcludeConfig|ExcludeLogs|" +
-        "ExcludeDirectoryBuild|ExcludeCi|SpecificationFiles|ExternalFiles|TasksAssembly|" +
-        "TasksDirectory)\\b)",
+        $@"\b{NukePrefix}(?=(?:{string.Join("|", msBuildPropertySuffixes)})\b)",
+        RegexOptions.Compiled);
+
+    private static readonly Regex packageReferenceVariablePattern = new(
+        @"^\$\((?<variable>[^)]+)\)$",
```
avidenic: This a great improvement and we do not need to worry about attribute ordering. Still skips things like `Version="$(Major).$(Minor)"` but I think this is a gap not worth filling and just add to docs.

src/Fallout.Migrate/CsprojRewriter.cs
```diff
@@ -1,84 +1,286 @@
...
+                    {
+                        var versionAttr = e.Attribute(VersionAttribute);
+                        if (versionAttr == null)
+                        {
+                            continue;
+                        }
+
+                        versionAttr.Value = $"$({FalloutVersionProperty})";
+                        edits++;
+                    }
```
avidenic: There is still small gap here and we might want to fix it with #492 - `$(FalloutVersion)` is rewritten even if there is no `<PropertyGroup>` in the file (e.g. it lives in `Directory.Build.props`). Or we can check for this in `InjectFalloutVersion`. So if `firstPropertyGroup == null` we add `<PropertyGroup>` on the root `<Project>` and add `<FalloutVersion>` to it instead of returning 0. The second option gives us a working solution but might break user preferences for having these kind of variables out of .csproj. We can note this in docs as well. Up to you =)

avidenic (approved):
Substantial improvements. There are still some decisions to be made though.

src/Fallout.Migrate/CsprojRewriter.cs
ITaluone: The problem with using `LoadOptions.PreserveWhitespace` is, that the serializer messes up indentation completely if preservered.. like... "ou.. the user has a specific indentation/formatting. Let's ignore"

src/Fallout.Migrate/CsprojRewriter.cs
ITaluone: Like I described above

src/Fallout.Migrate/CsprojRewriter.cs
ITaluone: I think I will add a new `PropertyGroup`.. Either way it is broken for the user.

I would fix all of the CPM quirks in #492 

[378eb2a](https://github.com/Fallout-build/Fallout/pull/478/commits/378eb2acc0fe313af11ea4412dfc32a8a4c3a764)

src/Fallout.Migrate/CsprojRewriter.cs
ITaluone: Ok [37d887a](https://github.com/Fallout-build/Fallout/pull/478/commits/37d887abe794a91d19b181a0004cb1eab0d7df09)

src/Fallout.Migrate/CsprojRewriter.cs
```diff
@@ -1,84 +1,287 @@
...
+    private const string VersionAttribute = "Version";
+    private const string FalloutVersionProperty = "FalloutVersion";
+    private const string PropertyGroupElement = "PropertyGroup";
 
     public static RewriteResult Rewrite(string original, string falloutVersion)
+    {
+        XDocument doc;
+        try
+        {
+            doc = XDocument.Parse(original, LoadOptions.None);
```
dennisdoomen: 🔧🤖 Parsing with `LoadOptions.None` and re-serializing via `doc.ToString(SaveOptions.None)` reformats the *entire* document, not just the touched regions. I verified this: blank lines between elements are stripped and all indentation gets normalized to the library's 2-space default, even for parts of the file the rewrite never touches. The previous regex approach left everything else byte-identical. Since `EditCount` implies a surgical/minimal-diff rewrite, this will produce large, unrelated reformatting diffs on any csproj with non-default indentation or blank-line separators. Worth considering `LoadOptions.PreserveWhitespace` (or documenting this as an accepted tradeoff).

src/Fallout.Migrate/CsprojRewriter.cs
```diff
@@ -1,84 +1,287 @@
...
+    {
+        if (!elements.Any(e =>
+                e.Name.LocalName == PackageReferenceElement && e.Parent != null &&
+                e.Attribute(VersionAttribute)?.Value == $"$({FalloutVersionProperty})") ||
+            elements.Any(e => string.Equals(e.Name.LocalName, FalloutVersionProperty, StringComparison.OrdinalIgnoreCase) && e.Parent != null))
+        {
+            return 0;
+        }
+
+        var firstPropertyGroup = elements.FirstOrDefault(e => e.Name.LocalName == PropertyGroupElement && e.Parent != null);
```
dennisdoomen: 🔧🤖 This picks the *first* `PropertyGroup` in document order without checking for a `Condition` attribute. If the first `PropertyGroup` in a project happens to be configuration-conditioned (common in multi-target/legacy projects), `FalloutVersion` would only be defined for that configuration, leaving `$(FalloutVersion)` unresolved elsewhere. Should this skip conditioned `PropertyGroup`s (or prefer an unconditioned one) when picking the injection target?

tests/Fallout.Migrate.Specs/CsprojRewriterSpecs.cs
```diff
@@ -77,13 +74,11 @@ public void ReturnsZeroEditsForUnchangedContent()
                              """;
 
         var result = CsprojRewriter.Rewrite(input, TestFalloutVersion);
-
-        result.EditCount.Should().Be(0);
-        result.Content.Should().Be(input);
+        await Verifier.Verify(result.Content);
```
dennisdoomen: ⛏🤖 `ReturnsZeroEditsForUnchangedContent` only verifies `result.Content` — it never asserts `result.EditCount.Should().Be(0)`, despite that being what the test name promises.

src/Fallout.Migrate/CsprojRewriter.cs
```diff
@@ -1,84 +1,287 @@
...
             edits++;
-            return "Fallout.";
-        });
+        }
+
+        return edits;
+    }
 
-        content = msBuildPropertyPattern.Replace(content, _ =>
+    private static int DecoupleVariables(List<XElement> elements, HashSet<string> variablesToBump)
```
dennisdoomen: ♻️🤖 `DecoupleVariables` (and `RenamePackagesAndBumpVersions`) run well past the project's 15-line method guideline. `DecoupleVariables` mixes two responsibilities — building the `packageByVariable` map and then deciding per-variable whether to decouple/bump — that could be split into two smaller, named methods for readability.

dennisdoomen (changes requested):
I'd like to up our game here, even for our migration code, as that is going to be existing for years to come. So I'm going to be a bit more critical on the code review level. Don't get me wrong. We love all these contributions, but Nuke's code hasn't been particularly well factored and documented. So bear with me. 

🔧 The PRs should explain what they are trying to do. Just a reference to an old (potentially convoluted) Nuke issue isn't enough anymore. 
📝 The new code in this PR is IMO too long, too cryptic and has too little documentation.

src/Fallout.Migrate/CsprojRewriter.cs
```diff
@@ -1,84 +1,287 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Text.RegularExpressions;
+using System.Xml.Linq;
 
 namespace Fallout.Migrate;
```
dennisdoomen: 🔧 Please add some XML docs to all members to explain their need and purpose

tests/Fallout.Migrate.Specs/CsprojRewriterSpecs.cs
```diff
@@ -178,8 +165,121 @@ public void LeavesOtherSystemPackagesAlone()
                              """;
 
         var result = CsprojRewriter.Rewrite(input, TestFalloutVersion);
+        await Verifier.Verify(result);
+    }
 
-        result.EditCount.Should().Be(0);
-        result.Content.Should().Be(input);
+    [Fact]
+    public async Task Recognizes_a_version_variable_prefixed_with_Nuke()
```
dennisdoomen: 🔧 For the new tests, I like to keep the expectation inline instead of using `Verify`. It's important to see the cause and effect as clear as possible.

docs/migration/from-nuke.md
```diff
@@ -113,6 +113,7 @@ The [`Fallout.Migrate.Analyzers`](https://www.nuget.org/packages/Fallout.Migrate
 
 ## Common gotchas
 
+- **Composed Version Formats**: `fallout-migrate` automatically bumps `PackageReference` versions when they are literal strings or simple single-variable references (e.g., `$(NukeVersion)`). It does **not** support migrating version strings in composed formats like `Version="$(Major).$(Minor)"`. You must update these variables manually to the new Fallout version.
```
dennisdoomen: 🤔 Why can't we support that as well?

dennisdoomen (changes requested):

IT-VBFK:
Ok.. I propose to start over with small focused steps

Ok @dennisdoomen?

dennisdoomen:
> Ok.. I propose to start over with small focused steps

Yeah, and base it on #507

IT-VBFK:
Cool
Will do so

I couldn't imagine how far your migration refactoring could go 😂😂

src/Fallout.Migrate/Steps/RewriteCsprojsStep.cs
```diff
@@ -19,7 +20,7 @@ internal sealed class RewriteCsprojsStep : IMigrationStep
     // a broken post-migrate build (#217). Tolerates extra attributes between Include and Version
     // (e.g. `PrivateAssets="all"`).
     private static readonly Regex nukePackageWithInlineVersionPattern = new(
-        @"(?<prefix><PackageReference\s+Include="")Nuke\.(?<name>[A-Z][A-Za-z0-9.]+)(?<between>""[^>]*?\s+Version="")[^""]+",
+        @"(?<prefix><PackageReference\s+Include="")Nuke\.(?<name>[A-Z][A-Za-z0-9.]+)(?<between>""[^>]*?\s+Version="")(?!\$\()[^""]+",
```
dennisdoomen, line 25: 🔧 The comment doesn't fully cover the new regex anymore.

src/Fallout.Migrate/Steps/RewriteCsprojsStep.cs
```diff
@@ -109,6 +115,32 @@ public static RewriteResult Rewrite(string original, string falloutVersion)
             return string.Empty;
         });
 
+        // Pass 4 — extract variables used in PackageReferences and bump their properties.
```
dennisdoomen: 🔧 I'd rather see this encapsulated into a new method so `Rewrite` stops from growing.

src/Fallout.Migrate/Steps/RewriteCsprojsStep.cs
```diff
@@ -109,6 +115,32 @@ public static RewriteResult Rewrite(string original, string falloutVersion)
...
+        var variablesToBump = packageReferenceVariablePattern.Matches(content)
+            .Select(m => m.Groups["variable"].Value)
+            .ToHashSet();
+
+        // Always include FalloutVersion as they are canonical.
```
dennisdoomen: ❓ What does that mean?

src/Fallout.Migrate/Steps/RewriteCsprojsStep.cs
```diff
@@ -109,6 +115,32 @@ public static RewriteResult Rewrite(string original, string falloutVersion)
             return string.Empty;
         });
 
+        // Pass 4 — extract variables used in PackageReferences and bump their properties.
+        var variablesToBump = packageReferenceVariablePattern.Matches(content)
```
dennisdoomen: ⛏️ Please only use `var` unless the type is visible

src/Fallout.Migrate/Steps/RewriteCsprojsStep.cs
```diff
@@ -109,6 +115,32 @@ public static RewriteResult Rewrite(string original, string falloutVersion)
...
+            .Select(m => m.Groups["variable"].Value)
+            .ToHashSet();
+
+        // Always include FalloutVersion as they are canonical.
+        variablesToBump.Add("FalloutVersion");
+
+        foreach (var variable in variablesToBump)
+        {
+            var pattern = $"(?<=<{variable}>)[^<]+(?=</{variable}>)";
```
dennisdoomen: ⛏️ I nice comment explaining what you're doing here so contributors don't have to analyze the entire regex would most definitely help.

dennisdoomen (changes requested):
Nice improvement with a little bit of rework

src/Fallout.Migrate/Steps/RewriteCsprojsStep.cs
```diff
@@ -28,13 +29,18 @@ internal sealed class RewriteCsprojsStep : IMigrationStep
     private static readonly Regex packageReferencePattern =
         new(@"(?<=\b(?:Include|Update|Remove)="")Nuke\.(?=[A-Z])", RegexOptions.Compiled);
 
+    // Detects MSBuild variables used in PackageReference Version attributes: Version="$(MyVar)"
+    private static readonly Regex packageReferenceVariablePattern = new(
+        @"<PackageReference\s+[^>]*?Version=""\$\((?<variable>[^)]+)\)""",
+        RegexOptions.Compiled);
```
dennisdoomen, line 45: 🔧🤖 `packageReferenceVariablePattern` matches *any* `<PackageReference Version="$(Var)">`, not just Nuke/Fallout ones. Pass 4 then unconditionally overwrites that property's value to `falloutVersion`. E.g. a project with `<PackageReference Include="System.Text.Json" Version="$(SystemTextJsonVersion)" />` would have `SystemTextJsonVersion` silently corrupted to the Fallout version. Should scope extraction to only variables referenced by `Fallout.*` package references (i.e. after the Pass 2 rename).

src/Fallout.Migrate/Steps/RewriteCsprojsStep.cs
dennisdoomen: 🔧🤖 The variable name is interpolated straight into a regex here without `Regex.Escape`. MSBuild/XML property names can contain regex metacharacters (e.g. `.`), risking wrong matches or a malformed regex.

tests/Fallout.Migrate.Specs/RewriteCsprojsStepSpecs.cs
```diff
@@ -183,4 +183,74 @@ public void LeavesOtherSystemPackagesAlone()
...
+                               </ItemGroup>
+                             </Project>
+                             """;
+
+        var result = RewriteCsprojsStep.Rewrite(input, TestFalloutVersion);
+
+        result.Content.Should().Contain("<UnreferencedVersion>10.3.49</UnreferencedVersion>")
+          .And.Contain("<CiProjectVersion>11.0.0</CiProjectVersion>")
+          .And.Contain("$(CiProjectVersion)", Exactly.Once());
+    }
```
dennisdoomen, line 315: 📝🤖 No regression test covers a non-Fallout package sharing a version variable — exactly the scenario that would catch the corruption described in the comment on `packageReferenceVariablePattern`. Worth adding one here.

dennisdoomen (commented):
Grouped review: three related findings on the new property-variable bump logic.

src/Fallout.Migrate/Steps/RewriteCsprojsStep.cs
IT-VBFK: Hmm.. removed it, still all tests passing. So.. nuke that :D

IT-VBFK:
Ok.. as I can see this is more or less the version like before, except the verify conversion

src/Fallout.Migrate/CsprojRewriter.cs
```diff
@@ -79,6 +88,34 @@ public static RewriteResult Rewrite(string original, string falloutVersion)
...
+
+        // Always include NukeVersion/FalloutVersion as they are canonical.
+        variablesToBump.Add("NukeVersion");
+        variablesToBump.Add("FalloutVersion");
+
+        foreach (var variable in variablesToBump)
+        {
+            var escapedVariable = Regex.Escape(variable);
+            var pattern = $@"(?<open><{escapedVariable}(?=\s|>)[^>]*>)(?<value>[^<]+)(?<close></{escapedVariable}>)";
+
```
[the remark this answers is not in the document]
dennisdoomen: 🔧🤖 Re-checked against the final shipped commit (39b1b12) — this is still live. I tested `<NukeVersion >9.0.0</NukeVersion >` (space before the closing `>`): the property gets renamed to `FalloutVersion` but its value stays `9.0.0` instead of bumping to the target version. The suggested named-group fix in the reply below never made it into the final regex-based rewrite (`RewriteCsprojsStep.cs`), which still uses the plain `(?<=<{variable}>)[^<]+(?=</{variable}>)` lookbehind/lookahead.

src/Fallout.Migrate/CsprojRewriter.cs
dennisdoomen: 🔧🤖 This gap is back, and worse, in the final code. I tested a project with an ambiguous shared variable but no `<PropertyGroup>` element at all: the Fallout `PackageReference` still gets redirected to `Version="$(FalloutVersion)"`, but no `<FalloutVersion>` property is ever injected (the `content.IndexOf("<PropertyGroup>", ...) >= 0` check just skips the injection silently). That leaves `$(FalloutVersion)` referenced but undefined — a broken csproj, worse than the original no-op gap.

src/Fallout.Migrate/Steps/RewriteCsprojsStep.cs
dennisdoomen: 🔧 Still true on the final commit (39b1b12). `nukePackageWithInlineVersionPattern`'s comment is unchanged and doesn't mention the `(?!\$\()` exclusion added to exclude property-based versions.

src/Fallout.Migrate/Steps/RewriteCsprojsStep.cs
```diff
@@ -19,7 +21,7 @@ internal sealed class RewriteCsprojsStep : IMigrationStep
     // a broken post-migrate build (#217). Tolerates extra attributes between Include and Version
     // (e.g. `PrivateAssets="all"`).
```
IT-VBFK, line 22:
```suggestion
    // (e.g. `PrivateAssets="all"`).
    // We don't mach MSBuild variables (`$(...)`) here, because they are handled below
```

src/Fallout.Migrate/Steps/RewriteCsprojsStep.cs
IT-VBFK: Ou.. sorry.. My local branch was way off and had that line. After rebase and conflict resolving this was gone.

ITaluone:
Can we move forward on that one?

And: can this go into 10.4?

dennisdoomen:
> Can we move forward on that one?

You've marked some of my (and my AI friend's) comments as resolved, but didn't answer nor fix anything. I've reopened those.

docs/migration/from-nuke.md
ITaluone: As we discussed in the [issue](https://github.com/Fallout-build/Fallout/issues/477#issuecomment-5019070919): let's discuss about such first in a new issue whether we want to support this.

ITaluone:
> > Can we move forward on that one?
> 
> You've marked some of my (and my AI friend's) comments as resolved, but didn't answer nor fix anything. I've reopened those.

Sorry that this went like that.. 

Added a new test case for the non-PropertyGroup case, and added whitespaces to the version property tags.

ChrisonSimtian:
im a bit lost with this one @IT-VBFK is this ready for review now? if we wanna get this in into 10.4 I'd say lets try to get this merged in the next few days :-)

IT-VBFK:
Yes
Ready for review

dennisdoomen (approved):

ChrisonSimtian (approved):

PR #479, opened 2026-07-11
Fix local run tests, when xunit.runner.console version is globally on…

### LINKED ISSUE #471: [Testing]: Constant `XunitConsolePackageVersion` set to 2.6.1 is probably never correct locally
### Usage Information

Latest main

### Description

`NuGetPackageResolverSpecs.XunitConsolePackageVersion` is set to "2.6.1", which is probably never correct nowadays.

The test `TestGetGlobalInstalledPackage()` fails at least for me when tests run locally, because some other projects in the past had the `xunit.runner.console` 2.9.3 dependency and this version is picked for this test.

This is unexpected, that other project's nuget dependencies have impact on tests on `Fallout`.

### Reproduction Steps

- Install `xunit.runner.console` 2.9.3 in any other project
- Run `NuGetPackageResolverSpecs.TestGetGlobalInstalledPackage` test
- see failing, because it expects 2.6.1 instead of 2.9.3

### Expected Behavior

Test fails

### Actual Behavior

Test should not be influences by other projects dependencies

### Regression?

No

### Known Workarounds

Use FA's `BeOneOf`
`result.Version.OriginalVersion.Should().BeOneOf(XunitConsolePackageVersion, "2.9.3");`

This will most likely be correct in the long term, because "2.9.3" is maybe the last xunit.runner.console v2 version...

This is the only way I can think of "fixing" this, despite of deleting the test...

What do you think @ChrisonSimtian @dennisdoomen ?

Yes

ITaluone:
… higher version

Fixes #471 

The problem is, that a globally installed xunit.runner.console nuget package from another project breaks this test.

### DIFF
--- a/tests/Fallout.Tooling.Specs/NuGetPackageResolverSpecs.cs
+++ b/tests/Fallout.Tooling.Specs/NuGetPackageResolverSpecs.cs
@@ -15,6 +15,9 @@ public class NuGetPackageResolverSpecs
 
     private const string XunitConsolePackageVersion = "2.6.1";
 
+    // This is the latest package version of xunit.runner.console (v2 is deprecated)
+    private const string LastXunitConsolePackageVersion = "2.9.3";
+
     [Theory]
     [InlineData("SpecK", true, true, "1.0.1-ci00055")]
     [InlineData("SpecK", false, true, "1.0.0")]
@@ -34,7 +37,7 @@ public class NuGetPackageResolverSpecs
         result.Should().NotBeNull();
         result.Id.Should().Be("xunit.runner.console");
         result.File.Name.Should().EndWith("nupkg");
-        result.Version.OriginalVersion.Should().Be(XunitConsolePackageVersion);
+        result.Version.OriginalVersion.Should().BeOneOf(XunitConsolePackageVersion, LastXunitConsolePackageVersion);
     }
 
     [Fact]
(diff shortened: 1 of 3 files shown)


tests/Fallout.Tooling.Specs/NuGetPackageResolverSpecs.cs
```diff
@@ -15,6 +15,9 @@ public class NuGetPackageResolverSpecs
 
     private const string XunitConsolePackageVersion = "2.6.1";
 
+    // This is the latest package version of xunit.runner.console (v2 is deprecated)
+    private const string LastXunitConsolePackageVersion = "2.9.3";
+
```
IT-VBFK:
```suggestion
```

tests/Fallout.Tooling.Specs/NuGetPackageResolverSpecs.cs
```diff
@@ -15,6 +15,9 @@ public class NuGetPackageResolverSpecs
 
     private const string XunitConsolePackageVersion = "2.6.1";
```
IT-VBFK:
```suggestion
    // This version have to match with
    // - Fallout.Tooling.Specs.csproj -> PackageDownload
    private const string XunitConsolePackageVersion = "2.9.3";
```

tests/Fallout.Tooling.Specs/NuGetPackageResolverSpecs.cs
```diff
@@ -34,7 +37,7 @@ public void TestGetGlobalInstalledPackage()
         result.Should().NotBeNull();
         result.Id.Should().Be("xunit.runner.console");
         result.File.Name.Should().EndWith("nupkg");
-        result.Version.OriginalVersion.Should().Be(XunitConsolePackageVersion);
+        result.Version.OriginalVersion.Should().BeOneOf(XunitConsolePackageVersion, LastXunitConsolePackageVersion);
```
IT-VBFK:
```suggestion
        result.Version.OriginalVersion.Should().Be(XunitConsolePackageVersion);
```

ChrisonSimtian (approved):

Issue #372, opened 2026-06-05
DRAFT: fall-out triggers compile errors


dennisdoomen:
Using `Fallout.Migrate` 10.3.49 on https://github.com/fluentassertions/fluentassertions resulted in

```
fallout-migrate - migrating: C:\Workspaces\fluentassertions

edit    Build/_build.csproj  (8 changes)
edit    Build/Build.cs  (18 changes)
edit    Build/CompressionExtensions.cs  (1 change)
edit    Build/Configuration.cs  (1 change)
edit    Build/CustomNpmTasks.cs  (6 changes)
edit    build.ps1  (2 changes)
edit    build.sh  (2 changes)

Files changed:   7
Edits made:      38
Directories:     0 renamed

Warnings:
  - Both .nuke/ and .fallout/ exist. Skipped rename; merge their contents manually.

Migration complete. Verify the build:  ./build.ps1   (or ./build.sh on unix)
Migration guide: https://fallout.build  (see #37 for the full guide)
```

After building, I got these errors

```
    C:\Workspaces\fluentassertions\build\_build.csproj : error NU1603: Warning As Error: _build depends on Fallout.Common (>= 10.1.0) but Fallout.Common 10.1.0 was not found. Fallout.Common 10.2.12 was resolved instead.
    C:\Workspaces\fluentassertions\build\_build.csproj : error NU1603: Warning As Error: _build depends on Fallout.Components (>= 10.1.0) but Fallout.Components 10.1.0 was not found. Fallout.Components 10.2.12 was resolved instead.
    C:\Workspaces\fluentassertions\build\_build.csproj : error NU1605:
      Warning As Error: Detected package downgrade: System.Security.Cryptography.Xml from 10.0.6 to 9.0.15. Reference the package
       directly from the project to select a different version.
       _build -> Fallout.Common 10.2.12 -> System.Security.Cryptography.Xml (>= 10.0.6)
       _build -> System.Security.Cryptography.Xml (>= 9.0.15 && < 10.0.0)
```

PackageGuard gave me the following errors

```
0>_build.csproj: Warning NU1603 : _build depends on Fallout.Common (>= 9.0.4) but Fallout.Common 9.0.4 was not found. Fallout.Common 10.2.12 was resolved instead.
0>_build.csproj: Warning NU1603 : _build depends on Fallout.Components (>= 9.0.4) but Fallout.Components 9.0.4 was not found. Fallout.Components 10.2.12 was resolved instead.
[...]

Issue #457, opened 2026-06-30
Environment Variables prepended with `FALLOUT_` don't work.


pkillboredom:
### Usage Information

10.3.49

### Description

Contrary to the information stated in the documentation for [Parameters](https://github.com/Fallout-build/Fallout/blob/f57634c5636bf5ba4fd0f49eaf0f8360697abcb3/docs/02-fundamentals/06-parameters.md?plain=1#L81), the code is still looking for `NUKE_`.

Problem Code:
https://github.com/Fallout-build/Fallout/blob/f57634c5636bf5ba4fd0f49eaf0f8360697abcb3/src/Fallout.Build/Execution/ParameterService.cs#L192

I can create a PR later to address this, but I am unsure if it should also remain compatible with `NUKE_`. LMK what you think.

### Reproduction Steps

1. Create Build Project
2. Add a string parameter.
3. Assert in target when parameter is null.
4. `export FALLOUT_MY_PARAMETER_NAME="hello"`
5. Run build, reach assertion.
6. `export NUKE_MY_PARAMETER_NAME="hello"`
7. Run build, no assertion.

### Expected Behavior

Environment variables prepended with `FALLOUT_` should be used for parameters as per documentation.

### Actual Behavior

See Repro.

### Regression?

Kind of. Technically?

### Could you help with a pull-request?

Yes

ITaluone:
Ok.. on your go

IT-VBFK:
> I can create a PR later to address this, but I am unsure if it should also remain compatible with NUKE_

Yes, of course. We are currently in the transition phase

ChrisonSimtian:
Hi @pkillboredom , happy for you to contribute and fix this via a pr :-) As for your question: i would say let's move forward and drop the NUKE_ (pun intended) if it complicates things too much. we can always slot breaking changes into the next major release

ITaluone:
@pkillboredom ping

PR #487, opened 2026-07-13
Rewrite fallout-migrate argument parsing and path handling

dennisdoomen:
Rewrite fallout-migrate's argument parsing and path handling to use Spectre.Console.Cli and Fallout.Common.IO.AbsolutePath. This is mostly to prepare for #372.

### What changed
- Replace the hand-rolled `Main` argument parser with a Spectre.Console.Cli `CommandApp<MigrateCommand>` (`CommandSettings` with `[CommandArgument]`/`[CommandOption]`, built-in `--help` generation).
- Replace raw `System.IO` path handling in `Migration.cs` with `AbsolutePath` (`/` combinator, `.FileExists()`/`.DirectoryExists()`, `.GetFiles(depth:)`, `.Move(ExistsPolicy.Fail)`, `.ReadAllText()`/`.WriteAllText(eofLineBreak: false)`, `.GetUnixRelativePathTo(...)`).
- Flip `Migration`, `Summary`, and `RewriteResult` to `internal`, matching the already-internal rewriter classes.
- Add `Spectre.Console.Cli` (0.49.1, matching `Spectre.Console`) to `Directory.Packages.props` and a `ProjectReference` to `Fallout.Utilities`.

### Why
`fallout-migrate` hand-rolled its own arg parsing and used raw string paths. This aligns it with the CLI framework already used for prompts (`Spectre.Console`) and the path type used throughout the rest of the codebase (`AbsolutePath`).

### Verification
- `dotnet build src/Fallout.Migrate` — clean.
- `dotnet test tests/Fallout.Migrate.Specs` — 22/22 passing, unchanged.
- Manual: `--help` renders via Spectre; `--dry-run` and a real run against a scratch NUKE-style fixture produced byte-identical rewrites and correctly renamed `.nuke/` → `.fallout/`.

Note: losing `/?` as a help alias (Spectre only recognizes `-h`/`--help`) is a minor, deliberate side effect of switching to Spectre's built-in help.

### DIFF
--- a/src/Fallout.Migrate/Migration.cs
+++ b/src/Fallout.Migrate/Migration.cs
@@ -1,21 +1,23 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Reflection;
+using Fallout.Common.IO;
 
 namespace Fallout.Migrate;
 
-public sealed class Migration
+internal sealed class Migration
 {
-    private readonly string _rootDirectory;
-    private readonly bool _dryRun;
-    private readonly TextWriter _log;
+    private readonly AbsolutePath rootDirectory;
+    private readonly bool dryRun;
+    private readonly TextWriter log;
 
-    public Migration(string rootDirectory, bool dryRun, TextWriter log)
+    public Migration(AbsolutePath rootDirectory, bool dryRun, TextWriter log)
     {
-        _rootDirectory = rootDirectory ?? throw new ArgumentNullException(nameof(rootDirectory));
-        _dryRun = dryRun;
-        _log = log ?? throw new ArgumentNullException(nameof(log));
+        this.rootDirectory = rootDirectory ?? throw new ArgumentNullException(nameof(rootDirectory));
+        this.dryRun = dryRun;
+        this.log = log ?? throw new ArgumentNullException(nameof(log));
     }
 
     public Summary Run()
@@ -34,7 +36,9 @@ public sealed class Migration
     {
         var falloutVersion = ResolveFalloutVersion();
         foreach (var path in EnumerateFiles("*.csproj"))
+        {
             ApplyRewrite(path, content => CsprojRewriter.Rewrite(content, falloutVersion), summary);
+        }
     }
 
     // Pinned into migrated `<PackageReference Include="Fallout.X" Version="..." />` lines.
@@ -45,17 +49,22 @@ public sealed class Migration
     // bogus pin like Version="LOCAL". Inlined to keep Fallout.Migrate dependency-free.
     private static string ResolveFalloutVersion()
     {
-        const string Fallback = "11.0.0";
+        const string fallback = "10.3.49";
 
         var informational = typeof(Migration).Assembly
             .GetCustomAttribute<AssemblyInformationalVersionAttribute>()
             ?.InformationalVersion;
-        if (string.IsNullOrEmpty(informational))
-            return Fallback;
 
-        var plusIndex = informational.IndexOf('+');
+        if (string.IsNullOrEmpty(informational))
+        {
+            return fallback;
+        }
+
+        int plusIndex = informational.IndexOf('+');
         if (plusIndex == -1)
-            return Fallback;
+        {
+            return fallback;
+        }
 
         return informational[..plusIndex];
     }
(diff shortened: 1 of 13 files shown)


src/Fallout.Migrate/Migration.cs
```diff
@@ -45,81 +49,105 @@ private void RewriteCsprojs(Summary summary)
     // bogus pin like Version="LOCAL". Inlined to keep Fallout.Migrate dependency-free.
     private static string ResolveFalloutVersion()
     {
-        const string Fallback = "11.0.0";
+        const string fallback = "11.0.0";
```
ChrisonSimtian: i don't think this is correct anymore. lets use one of the 10.x?

ChrisonSimtian (approved):

src/Fallout.Migrate/Program.cs
```diff
@@ -1,38 +1,103 @@
...
 
-        var rootDirectory = ResolveRootDirectory(rootArg);
-        if (rootDirectory == null)
+[Description("""
+             Migrate a NUKE consumer repo to Fallout.
+
+             What it does:
+               - Rewrites Nuke.* PackageReferences and MSBuild properties in .csproj
+               - Rewrites `using Nuke.*` directives and qualified type references in .cs
+               - Rewrites `dotnet nuke` -> `dotnet fallout` and legacy NUKE_* env vars
```
ITaluone: Should we _add_ the new `FALLOUT_` env vars rather than rename the legacy ones? Since this (in the 10.x stage) is a drop-in replacement this would be a breaking change?

Or do we simply not care?

ITaluone (approved):

src/Fallout.Migrate/Program.cs
dennisdoomen: I might be missing something. What do you mean with that?

avidenic:
@dennisdoomen great contribution. Something I noticed as well, though I did not notice you PR before opening an issue. Would you be so kind to add a conformation prompt as per #491 ?

src/Fallout.Migrate/Program.cs
ITaluone: Ou.. maybe I am missing something? I thought about this env vars (the Fallout handling part) https://github.com/Fallout-build/Fallout/issues/457

But if this means the `NUKE_` variables in `build(.sh|ps1)` then I was wrong

dennisdoomen:
> Would you be so kind to add a conformation prompt

What do you mean with that?

src/Fallout.Migrate/Program.cs
dennisdoomen: But what does that have to do with the migration tool?

src/Fallout.Migrate/Migration.cs
```diff
@@ -45,81 +49,105 @@ private void RewriteCsprojs(Summary summary)
...
+
             return;
         }
 
         Log($"rename  {RelativePath(legacy)} -> {RelativePath(canonical)}");
-        if (!_dryRun)
-            Directory.Move(legacy, canonical);
+        if (!dryRun)
+        {
+            legacy.Move(canonical, ExistsPolicy.Fail);
```
avidenic: This swap changes how the rename fails, not just the API it calls. `Directory.Move` did the whole thing as one OS operation: atomic, all or nothing. `AbsolutePath.Move` routes to `MoveDirectory`, which creates `.fallout/`, moves each subdirectory and file one at a time, then deletes the source. So if one of those moves fails partway, say a locked file under `.nuke/temp` or a permissions problem, you're left half-migrated: part of the tree under `.fallout/`, the rest still under `.nuke/`, and `.nuke/` never deleted. No rollback.

The happy path is fine and the integration test covers it, so this is only about what happens when it breaks. But this thing rewrites someone else's repo, so the rename is exactly where I'd want to keep the atomic behavior.

Was the switch deliberate, maybe to pick up cross-volume moves that `Directory.Move` can't do? If so, a one-line comment saying that would save the next reader the same double take. If not, `Directory.Move` is the safer call here.

src/Fallout.Migrate/Migration.cs
```diff
@@ -45,81 +49,105 @@ private void RewriteCsprojs(Summary summary)
     // bogus pin like Version="LOCAL". Inlined to keep Fallout.Migrate dependency-free.
     private static string ResolveFalloutVersion()
     {
-        const string Fallback = "11.0.0";
+        const string fallback = "10.0.0";
```
avidenic: The tool ships on `2026.1.0-preview` (version.json), so a `10.0.0` fallback looks odd at first. I can see the reasoning: `11.0.0` went stale once v11 was retired, and `10.0.0` is the newest `Fallout.X` that actually exists on nuget.org, which makes it the safest thing to fall back to. But it's also the one value that never matches the tool's own major, so anyone who assumes the fallback tracks the current line will stop and wonder.

Could you add a short comment on why it's `10` (restorable from nuget.org, not the current version)? This only fires when the assembly informational version is missing, so it's low stakes, just easy to misread.

src/Fallout.Migrate/Program.cs
IT-VBFK: Well... migrating the env vars while not listening to them can lead to unintended behavior.
That's all I wanted to express 🙃

avidenic:
> > Would you be so kind to add a conformation prompt
> 
> What do you mean with that?

Sorry, I mean the second half #491 . `fallout-migrate` rewrites files in target folder in place. I was thinking that having a confirmation prompt before committing to changes might be a good idea. Just a safety belt - so user confirms that is the intended folder. 

e.g.:
```
About to migrate /tmp/repoB in place. Continue? [y/N]
```

I think this is built into `Spectre.Console.Cli` with `AnsiConsole.Confirm(...)`. 

A `-y` or `--yes` would then keep non-interactive runs working as well.

src/Fallout.Migrate/Program.cs
dennisdoomen: I guess that's a run-time problem, not a migration problem. You're not using those environment variables in your build script, but passing them from your own environment.

src/Fallout.Migrate/Migration.cs
dennisdoomen: I see your point. Although nothing in this migration tool is atomic, it seems to be a bit more efficient. Changed it back.

src/Fallout.Migrate/Migration.cs
dennisdoomen: The latest version of the tool is 10.3.49. I'm going to be looking at that separately as I'm fixing #372.

dennisdoomen:
> > > Would you be so kind to add a conformation prompt
> > 
> > 
> > What do you mean with that?
> 
> Sorry, I mean the second half #491 . `fallout-migrate` rewrites files in target folder in place. I was thinking that having a confirmation prompt before committing to changes might be a good idea. Just a safety belt - so user confirms that is the intended folder.
> 
> e.g.:
> 
> ```
> About to migrate /tmp/repoB in place. Continue? [y/N]
> ```
> 
> I think this is built into `Spectre.Console.Cli` with `AnsiConsole.Confirm(...)`.
> 
> A `-y` or `--yes` would then keep non-interactive runs working as well.

Added that to #372

PR #493, opened 2026-07-15
Escape single quotes correctly in YAML CI writers

### LINKED ISSUE #483: CI writers emit YAML names via SingleQuote(), which backslash-escapes (invalid YAML for embedded quotes)
## Problem

`StringExtensions.SingleQuote()` escapes an embedded single quote with a backslash (`\'`). That is invalid inside a YAML single-quoted scalar — YAML escapes a quote by doubling it (`''`). The GitHub Actions and Azure Pipelines writers quote generated names/values through `.SingleQuote()`, so a value containing an apostrophe emits invalid YAML. It's latent today only because those values are machine-generated from controlled inputs (target names, artifact paths, image ids).

`SingleQuoteYaml()` — the YAML-correct primitive (doubles the quote) — was added next to `SingleQuote()` in #475. `SingleQuote()` must stay as-is for its shell / telemetry / error-message / tool-path / TeamCity callers, where doubling would be wrong.

## Outcome

The YAML writers quote via `SingleQuoteYaml()`. Generated YAML (and GitHub-expression strings) stay valid regardless of the characters in a quoted value.

## Call sites to migrate

Reference example already done: `GitHubActionsCustomStep.cs` quotes its name via `Name.SingleQuoteYaml()` — copy that pattern.

GitHub Actions (`src/Fallout.Common/CI/GitHubActions/Configuration/`):
- [ ] `GitHubActionsArtifactStep.cs:15` — step `name:`
- [ ] `GitHubActionsCacheStep.cs:17` — step `name:`
- [ ] `GitHubActionsCacheStep.cs:27` — the `hashFiles('…')` key args (GitHub-expression string literals; expression syntax also doubles the quote, so `SingleQuoteYaml` is correct here too)
- [ ] `GitHubActionsRunStep.cs:31` — step `name:`

Azure Pipelines (`src/Fallout.Common/CI/AzurePipelines/Configuration/`):
- [ ] `AzurePipelinesJob.cs:22` — `displayName:`
- [ ] `AzurePipelinesJob.cs:29` — `vmImage:` — ⚠️ this calls `.SingleQuote().SingleQuote()` (quotes twice, unlike `AzurePipelinesStage.cs:28` which quotes once). Looks like a pre-existing quirk; check the intended output before converting, don't just swap the method name.
- [ ] `AzurePipelinesPublishStep.cs:16` — `displayName:`
- [ ] `AzurePipelinesPublishStep.cs:20` — `pathToPublish:`
- [ ] `AzurePipelinesCacheStep.cs:30` — `displayName:`
- [ ] `AzurePipelinesCmdStep.cs:20` — `displayName:`
- [ ] `AzurePipelinesStage.cs:21` — `displayName:`
- [ ] `AzurePipelinesStage.cs:28` — `vmImage:`
- [ ] `AzurePipelinesDownloadStep.cs:20` — `downloadPath:`
- [ ] `AzurePipelines.cs:178` — `runTitle` pair

Leave unchanged (not YAML — doubling would be wrong): the `SingleQuote()` callers in `FalloutBuild.Statics.cs`, `Execution/ExecutionPlanner.cs`, `Telemetry/Telemetry.cs`, `IO/AbsolutePath.cs`, `Text/String.Quoting.cs`, `NuGetToolPathResolver.cs`, `Tools/GitHub/GitHubTasks.cs`, `Tools/OctoVersion/OctoVersionTasks.cs`, and `CI/TeamCity/TeamCity.cs` (TeamCity service-message format has its own escaping).

## Acceptance criteria

- Each YAML/expression call site above uses `SingleQuoteYaml()` (or, for `AzurePipelinesJob.cs:29`, whatever the correct single application is once the double-quote is understood).
- Non-YAML `SingleQuote()` callers left unchanged.
- A regression spec with an apostrophe in a generated name for at least one GitHub Actions and one Azure Pipelines writer.
- No snapshot churn on existing controlled-input cases (output is identical when the value contains no quote).

Ref: `SingleQuoteYaml` in `src/Fallout.Utilities/Text/String.Quoting.cs` (#475).

YousifSala7:
Fixes #483

**Description:**
This PR updates the YAML CI writers to use `SingleQuoteYaml()` instead of `SingleQuote()`, ensuring that embedded apostrophes are properly escaped by doubling the quote (e.g., `''`) rather than using a backslash `\'` which results in invalid YAML.

**Changes:**
- Migrated all requested call sites in `GitHubActions` and `AzurePipelines` configuration files to `SingleQuoteYaml()`.
- Addressed the pre-existing quirk in `AzurePipelinesJob.cs` (`vmImage`). Removed the double wrapping (`.SingleQuote().SingleQuote()`) and replaced it with a single `SingleQuoteYaml()` call, which produces the correct intended output.
- Left the non-YAML `SingleQuote()` callers unchanged as requested.
- Added regression specs in `AzurePipelinesJobSpecs` and `GitHubActionsArtifactStepSpecs` to verify apostrophe escaping and correct `vmImage` rendering without snapshot churn.

### DIFF
--- a/tests/Fallout.Common.Specs/CI/AzurePipelinesJobSpecs.cs
+++ b/tests/Fallout.Common.Specs/CI/AzurePipelinesJobSpecs.cs
@@ -1,0 +1,44 @@
+﻿using System.IO;
+using System.Threading.Tasks;
+using Fallout.Common.CI.AzurePipelines;
+using Fallout.Common.CI.AzurePipelines.Configuration;
+using Fallout.Common.Utilities;
+using VerifyXunit;
+using Xunit;
+
+namespace Fallout.Common.Specs.CI;
+
+public class AzurePipelinesJobSpecs
+{
+    private static Task Verify(AzurePipelinesJob job)
+    {
+        var stream = new MemoryStream();
+        var writer = new StreamWriter(stream, leaveOpen: true);
+        job.Write(new CustomFileWriter(writer, indentationFactor: 2, commentPrefix: "#"));
+        writer.Flush();
+
+        stream.Seek(offset: 0, SeekOrigin.Begin);
+        return Verifier.Verify(new StreamReader(stream).ReadToEnd());
+    }
+
+    [Fact]
+    public Task DisplayName_with_an_apostrophe_is_yaml_escaped()
+        => Verify(new AzurePipelinesJob
+                  {
+                      Name = "Compile",
+                      DisplayName = "Yousif's Job",
+                      Dependencies = new AzurePipelinesJob[0],
+                      Steps = new AzurePipelinesStep[0]
+                  });
+
+    [Fact]
+    public Task VmImage_is_quoted_correctly_without_double_wrapping_quirk()
+        => Verify(new AzurePipelinesJob
+                  {
+                      Name = "Compile",
+                      DisplayName = "Compile Job",
+                      Dependencies = new AzurePipelinesJob[0],
+                      Steps = new AzurePipelinesStep[0],
+                      Image = AzurePipelinesImage.Ubuntu2204
+                  });
+}
--- a/tests/Fallout.Common.Specs/CI/AzurePipelinesJobSpecs.cs
+++ b/tests/Fallout.Common.Specs/CI/AzurePipelinesJobSpecs.cs
@@ -2,28 +2,15 @@
 using System.Threading.Tasks;
 using Fallout.Common.CI.AzurePipelines;
 using Fallout.Common.CI.AzurePipelines.Configuration;
-using Fallout.Common.Utilities;
-using VerifyXunit;
 using Xunit;
 
 namespace Fallout.Common.Specs.CI;
 
 public class AzurePipelinesJobSpecs
 {
-    private static Task Verify(AzurePipelinesJob job)
-    {
-        var stream = new MemoryStream();
-        var writer = new StreamWriter(stream, leaveOpen: true);
-        job.Write(new CustomFileWriter(writer, indentationFactor: 2, commentPrefix: "#"));
-        writer.Flush();
-
-        stream.Seek(offset: 0, SeekOrigin.Begin);
-        return Verifier.Verify(new StreamReader(stream).ReadToEnd());
-    }
-
     [Fact]
     public Task DisplayName_with_an_apostrophe_is_yaml_escaped()
-        => Verify(new AzurePipelinesJob
+        => ConfigurationEntityVerifier.Verify(new AzurePipelinesJob
                   {
                       Name = "Compile",
                       DisplayName = "Bob's Job",
@@ -33,7 +20,7 @@ public class AzurePipelinesJobSpecs
 
     [Fact]
     public Task VmImage_is_quoted_correctly_without_double_wrapping_quirk()
-        => Verify(new AzurePipelinesJob
+        => ConfigurationEntityVerifier.Verify(new AzurePipelinesJob
                   {
                       Name = "Compile",
                       DisplayName = "Compile Job",
--- a/tests/Fallout.Common.Specs/CI/AzurePipelinesJobSpecs.cs
+++ b/tests/Fallout.Common.Specs/CI/AzurePipelinesJobSpecs.cs
@@ -26,7 +26,7 @@ public class AzurePipelinesJobSpecs
         => Verify(new AzurePipelinesJob
                   {
                       Name = "Compile",
-                      DisplayName = "Yousif's Job",
+                      DisplayName = "Bob's Job",
                       Dependencies = new AzurePipelinesJob[0],
                       Steps = new AzurePipelinesStep[0]
                   });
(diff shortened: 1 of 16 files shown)


tests/Fallout.Common.Specs/CI/GitHubActionsArtifactStepSpecs.Name_with_an_apostrophe_is_yaml_escaped.verified.txt
```diff
@@ -0,0 +1,5 @@
+﻿- name: 'Publish: Yousif''s Build'
+  uses: actions/upload-artifact@v5
+  with:
+    name: Yousif's Build
```
ITaluone: This seems not right. Can you recheck?

ITaluone (changes requested):

tests/Fallout.Common.Specs/CI/GitHubActionsArtifactStepSpecs.Name_with_an_apostrophe_is_yaml_escaped.verified.txt
YousifSala7: Hi @ITaluone, thanks for the review!
I rechecked this, The with.name field is intentionally emitted as a plain YAML scalar, and an apostrophe is valid there without escaping. The change in this PR is scoped to YAML values that are explicitly single-quoted (and the relevant expression call sites), per the issue and acceptance criteria, I also left with.name unchanged because it never used SingleQuote() originally.
If I misunderstood your comment and you were referring to something else, please let me know and I'll be happy to update it.

*(Also, apologies if using my own name in the spec was a bit unusual—I can easily rename it to `Bob's Build` to match the other tests if you prefer!)*

ITaluone: Yeah.. that was my thought :) If the single quote there doesn't make the yaml invalid: all good

tests/Fallout.Common.Specs/CI/AzurePipelinesJobSpecs.cs
```diff
@@ -0,0 +1,44 @@
...
+        stream.Seek(offset: 0, SeekOrigin.Begin);
+        return Verifier.Verify(new StreamReader(stream).ReadToEnd());
+    }
+
+    [Fact]
+    public Task DisplayName_with_an_apostrophe_is_yaml_escaped()
+        => Verify(new AzurePipelinesJob
+                  {
+                      Name = "Compile",
+                      DisplayName = "Yousif's Job",
```
ITaluone: **non-blocking:** change the name to some example name :)

ITaluone (approved):
Thanks for contributing this!

only this little nitpick :)

tests/Fallout.Common.Specs/CI/GitHubActionsArtifactStepSpecs.Name_with_an_apostrophe_is_yaml_escaped.verified.txt
ITaluone: huh.. as I am keep looking at this: Why should we handle the first `Name` with `SingleQuoteYaml()` but leave the second `Name`  alone? This makes no sense in my head 🤔

@avidenic Do you have any thoughts on that one? 

Edit: Scratch that! I just realized why the single quotes are needed 🤦‍♂️ (the `:` is the villain here, not the `'`)

tests/Fallout.Common.Specs/CI/GitHubActionsArtifactStepSpecs.cs
```diff
@@ -0,0 +1,26 @@
...
+    private static Task Verify(GitHubActionsArtifactStep step)
+    {
+        var stream = new MemoryStream();
+        var writer = new StreamWriter(stream, leaveOpen: true);
+        step.Write(new CustomFileWriter(writer, indentationFactor: 2, commentPrefix: "#"));
+        writer.Flush();
+
+        stream.Seek(offset: 0, SeekOrigin.Begin);
+        return Verifier.Verify(new StreamReader(stream).ReadToEnd());
+    }
```
avidenic: The method `Verify(...)` is being repeated almost verbatim in `AzurePipelinesJobSpecs`. I suggest extracting it to a helper so it can be shared e.g.:
```csharp
    public static Task Verify(ConfigurationEntity entity)
    {
        var stream = new MemoryStream();
        var writer = new StreamWriter(stream, leaveOpen: true);
        entity.Write(new CustomFileWriter(writer, indentationFactor: 2, commentPrefix: "#"));
        writer.Flush();

        stream.Seek(offset: 0, SeekOrigin.Begin);
        return Verifier.Verify(new StreamReader(stream).ReadToEnd());
    }
```

avidenic (approved):
Good work, only a small nitpick suggestion.

tests/Fallout.Common.Specs/CI/AzurePipelinesJobSpecs.cs
YousifSala7: Done!

tests/Fallout.Common.Specs/CI/GitHubActionsArtifactStepSpecs.cs
YousifSala7: Done! Thanks for the suggestion.

PR #494, opened 2026-07-16
Fix typo in tool installation command

GeorgDangl:
Small one😀


ITaluone:
Hmm.. There are a few more places where these two variants are mentioned (like in the README, but there calling `Fallout.GlobalTool` the deprecated one..

Maybe it should go towards `Fallout.GlobalTools` but renamed too quick?

@ChrisonSimtian I have this confusion all the time.. What is the (for the long term) correct/supported one?

dennisdoomen:
Thanks!

PR #497, opened 2026-07-17
Consolidate and rename CI workflows

ChrisonSimtian:
Cleans up `.github/workflows/` — clearer names, one fewer file, and fixes the long-broken README build-status badges. **No behaviour change** to CI triggers or the required check.

## Workflows

Two generated build workflows (from `build/Build.CI.GitHubActions.cs`) replace the three per-OS files:
- **`build.yml`** — Linux PR gate for PRs to `main`/`release/*`/`support/*`. Job stays named `ubuntu-latest`, so the **required status check and branch protection are untouched**.
- **`build-cross-platform.yml`** — macOS + Windows in one workflow (a job per OS, via the attribute's `params` image list), gated to release intent — PRs to `release/*`/`support/*` and `v*` tags only. Replaces `macos-latest.yml` + `windows-latest.yml`.

Hand-written workflows renamed for clarity:
- `ubuntu-latest-docs.yml` → **`build-skip.yml`** — the no-op that reports the `ubuntu-latest` check on docs-only PRs (it does not build docs; the old name implied it did).
- `preview.yml` → **`publish-packages-preview.yml`**
- `release.yml` → **`publish-packages-release.yml`** — `ReleaseWorkflow` const in `Build.CI.GitHubActions.cs` updated to match (it gates `CreateGitHubRelease`).

## README build-status fix

Badges had been broken for a while: they used `?branch=main` on `pull_request`-only workflows (always renders "no status"). Now the headline CI badge tracks `publish-packages-preview` (runs on every `main` push), and table badges use trigger-matched query params. Owner stays `Fallout-build/Fallout`.

## Why it's safe
- **Required check unchanged** — branch protection keys on the job name `ubuntu-latest`, which `build.yml` still emits.
- **Triggers identical** to before — rename + consolidation, not a behaviour change.
- ADRs left as point-in-time records; only living docs updated.

## Verification
- `dotnet fallout Test Pack` locally → **108 tests pass, 0 fail**; all 24 packages pack.
- Merged to the fork's `main` and exercised end-to-end: `build.yml` (real Test+Pack) and `build-skip.yml` both green; `publish-packages-preview` build/test/pack green (its publish step 403s only because a fork can't push to this org's package feed — pre-existing, resolves here on upstream).
- Both generated workflows freshly regenerated from the attributes (not hand-edited); all 5 workflow YAMLs parse clean.


ChrisonSimtian:
@dennisdoomen doing some tidy up, getting the naming aligned and making it more obvious what which workflow does.
I'll do a follow up PR at some point to get the remaining workflows into Fallout rather than having handcrafted ymls

dennisdoomen:
🏕️🤖 Stale reference (pre-existing file, not touched by this PR — flagging for a follow-up, not blocking): `build/Build.cs` lines 48 and 195 still mention `.github/workflows/preview.yml` and `release.yml`. This PR renamed those to `publish-packages-preview.yml` and `publish-packages-release.yml` everywhere else, but `Build.cs` wasnt part of the diff so these two comments were missed.

dennisdoomen (approved):

ITaluone (approved):
NIce!

PR #501, opened 2026-07-17
Use Microsoft.Windows.SDK.BuildTools package to get signtool binary

### LINKED ISSUE #1186: "SignToolTasks.SignToolPath" can't find the correct path for signtool.exe
### Usage Information

Nuke.Common 7.0.2 / Microsoft.NET 4.8.1 / Windows 10

### Description

When trying to run "SignToolTasks.SignTool()", I have the following error:

> 
> System.ArgumentException: Expected condition to be true (Parameter 'toolPath != null')
> at Nuke.Common.Assert.True(Boolean condition, String message, String argumentExpression) in /_/source/Nuke.Common/Assert.cs:line 34

"SignToolTasks.SignToolPath" looks for an old array of paths that don't correspond anymore with the latest windows kits released.
(current code: https://github.com/nuke-build/nuke/blob/develop/source/Nuke.Common/Tools/SignTool/SignToolSettings.cs)

For instance, I have "signtool.exe" installed on the path: "C:\Program Files (x86)\Windows Kits\10\bin\10.0.22621.0\x64\signtool.exe" (and the env variable "SIGNTOOL_EXE" not set)

### Reproduction Steps

Install any Windows Kit 10.*

### Expected Behavior

signtool should work without problems

### Actual Behavior

An exception is thrown

### Regression?

_No response_

### Known Workarounds

Set manually the env variable "SIGNTOOL_EXE" with the correct path.

Yes

IT-VBFK:
From: https://github.com/nuke-build/nuke/pull/1558

Fixes: https://github.com/nuke-build/nuke/issues/1186

Thanks: @gpailler

### DIFF
--- a/src/Fallout.Common/Tools/SignTool/SignTool.Generated.cs
+++ b/src/Fallout.Common/Tools/SignTool/SignTool.Generated.cs
@@ -18,9 +18,12 @@ namespace Fallout.Common.Tools.SignTool;
 
 /// <summary><p>Sign Tool is a command-line tool that digitally signs files, verifies signatures in files, and time-stamps files.</p><p>For more details, visit the <a href="https://docs.microsoft.com/en-us/dotnet/framework/tools/signtool-exe">official website</a>.</p></summary>
 [ExcludeFromCodeCoverage]
-public partial class SignToolTasks : ToolTasks
+[NuGetTool(Id = PackageId, Executable = PackageExecutable)]
+public partial class SignToolTasks : ToolTasks, IRequireNuGetPackage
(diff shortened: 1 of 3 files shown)


ChrisonSimtian (approved):
cheeky :-)

avidenic:
Looks good. Worth noting somewhere that we now depend on `Microsoft.Windows.SDK.BuildTools` in case someone depends on machine installed signtools. Unlikely as it was a stale reference, but you never know.

avidenic (approved):

PR #503, opened 2026-07-17
Add tool wrapper definition for `Stryker`

IT-VBFK:
From here: https://github.com/nuke-build/nuke/pull/1259

Thanks @candoumbe

### DIFF
--- a/src/Fallout.Common/Tools/Stryker/Stryker.Generated.cs
+++ b/src/Fallout.Common/Tools/Stryker/Stryker.Generated.cs
@@ -1,0 +1,537 @@
+// Generated from https://github.com/Fallout-build/Fallout/blob/main/src/Fallout.Common/Tools/Stryker/Stryker.json
+
+using Fallout.Common;
+using Fallout.Common.Tooling;
+using Fallout.Common.Tools;
+using Fallout.Common.Utilities.Collections;
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.Json.Serialization;
+
+namespace Fallout.Common.Tools.Stryker;
+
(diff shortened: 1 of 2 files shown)


dennisdoomen (commented):
⛏🤖 **Nitpick**: `BreakAt` is typed `uint`, but `ThresholdHigh`/`ThresholdLow` are typed `short`, even though all three represent the same 0-100 mutation-score concept and map to the same `int?` type upstream in Stryker (`ThresholdBreakInput`, `ThresholdHighInput`, `ThresholdLowInput`). Suggest using one consistent numeric type across all three.

❓🤖 **Question**: the wrapper omits several newer Stryker CLI options that exist in current `stryker-net` master (`CommandLineConfigReader.cs`): `with-baseline`, `test-runner`, `diag`, `azure-fileshare-sas`, `s3-bucket-name`, `s3-endpoint`, `s3-region`. Was this a deliberate scope match with the upstream NUKE PR (nuke-build/nuke#1259), or should these be added now?

IT-VBFK:
Thanks for review

I "just" took the definitions form the original PR

Will add them (or let them add 🤖)

IT-VBFK:
> ⛏🤖 Nitpick: BreakAt is typed uint, but ThresholdHigh/ThresholdLow are typed short, even though all three represent the same 0-100 mutation-score concept and map to the same int? type upstream in Stryker (ThresholdBreakInput, ThresholdHighInput, ThresholdLowInput). Suggest using one consistent numeric type across all three.


Used `uint` since other numbers also use `uint`

> ❓🤖 Question: the wrapper omits several newer Stryker CLI options that exist in current stryker-net master (CommandLineConfigReader.cs): with-baseline, test-runner, diag, azure-fileshare-sas, s3-bucket-name, s3-endpoint, s3-region. Was this a deliberate scope match with the upstream NUKE PR (https://github.com/nuke-build/nuke/pull/1259), or should these be added now?

Added them now since there were easy :)

ChrisonSimtian (approved):
lgtm but lets slow down a bit on adding new tools to the mix :-)
I'd longterm like to break those out into their own little spaces so that people can expand Fallout as they need, without making the core too heavy ;-)

PR #504, opened 2026-07-17
Add tool definition for `cmake`

IT-VBFK:
From here: https://github.com/nuke-build/nuke/pull/1006

Thanks @MoodieG

### DIFF
--- a/src/Fallout.Common/Tools/CMake/CMake.Generated.cs
+++ b/src/Fallout.Common/Tools/CMake/CMake.Generated.cs
@@ -1,0 +1,645 @@
+// Generated from https://github.com/Fallout-build/Fallout/blob/main/src/Fallout.Common/Tools/CMake/CMake.json
+
+using Fallout.Common;
+using Fallout.Common.Tooling;
+using Fallout.Common.Tools;
+using Fallout.Common.Utilities.Collections;
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.Json.Serialization;
+
+namespace Fallout.Common.Tools.CMake;
+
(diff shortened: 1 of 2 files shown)


dennisdoomen:
Should there not be a `.generated.cs` file as well?

IT-VBFK:
Hmmm... the contribution guidelines stating to explicity _not_ to commit generated tool files (since they are done once before a release).

If this should be different here vs. NUKE this may should be mentioned.

fredericvaillancourt:
Adding myself to the discussion to get the tool generated answer. Generated my files but it does change the generated from (url) ... several places in every tools ... In fact the current url is from a form and not even the official. Personally I don't like generated files in source control except when they are quite difficult to generate. Waiting to get a definitive answer before I make my PR.

IT-VBFK:
I vote for removing them all, because fixing errors in tools wrappers has to go into the json anyway and would be overwritten when not done).

I imagine following workflow
- add the `GenerateToolSettings` as additional release pipeline executed target 
- these files (intermediately generated) compiles into the production assetes (dll/nuget package).

You can generate the wrappers locally if you searching/fixing a wrapper bug.

Edit:
Ok.. I found a few tool wrappers who wraps a nuget package. These wrappers use partial classes and reference a generated value in this additional partial class.

ChrisonSimtian:
look there is a series of things on this topic, but tldr is: commit the `*.Generated.cs` for now please.
We could probably build those as part of the build but then thats build time on the CI that might as well sit on your machine locally ;-)
Not sure why and how but the convention and agents.md are wrong there. Might be worth to do a follow up and fix that

ChrisonSimtian:
oh and since this one is a tad grave digging (4 year old PR), lets make sure that the cmake stuff is still up2date please, yes? :-)
I personally dont use cmake so I rely a bit on someone who does to test this and give their opinion

IT-VBFK:
I already asked AI to add missing parameters and check the rest. Since I don't use cmake either (I'm a 100% .NET guy) we have to rely on actual c/c++ devs I suppose 🫣

dennisdoomen (approved):

PR #508, opened 2026-07-18
Style fallout-migrate's banner and summary output with AnsiConsole

dennisdoomen:
Stacked on #507. Replaces the plain Console.WriteLine calls in MigrateCommand's banner and summary printing with Spectre.Console's AnsiConsole.

image

### What changed
- Bold banner text, yellow dry-run notice
- Bold counts, yellow warnings, green completion message, dimmed guide link in the summary
- Interpolated values (paths, warning text) go through MarkupLineInterpolated so user-controlled content can't be misread as markup

Left untouched:
- Console.Error.WriteLine for the two error messages -- stderr semantics matter for piping/exit codes
- The TextWriter-based per-file/per-rename logging in Steps/Common -- stays plain text so MigrationIntegrationSpecs can keep capturing it via TextWriter.Null without a console

No behavior change; internal-only, no public API touched.

### Verification
dotnet test tests/Fallout.Migrate.Specs -- 22/22 pass.

### DIFF
--- a/src/Fallout.Migrate/MigrateCommand.cs
+++ b/src/Fallout.Migrate/MigrateCommand.cs
@@ -6,6 +6,7 @@ using Fallout.Common;
 using Fallout.Common.IO;
 using Fallout.Migrate.Common;
 using JetBrains.Annotations;
+using Spectre.Console;
 using Spectre.Console.Cli;
 
 namespace Fallout.Migrate;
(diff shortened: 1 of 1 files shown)


ChrisonSimtian:
mind adding a screenshot to the PR for future reference?

ChrisonSimtian (approved):

PR #509, opened 2026-07-18
Improve fallout-migrate: NuGet-scoped version pin, TFM warning/bump, confirmation prompt, async pipeline

dennisdoomen:
## Summary
A set of focused improvements to `fallout-migrate`: pin the migrated `Fallout.Common` reference to the latest NuGet release instead of the running tool's own version, warn and optionally bump an outdated build-project TFM, ask for confirmation before writing files, and convert the migration pipeline to async.

image

## Changes

**Bump Fallout.Migrate to the 10.x version line** — version bump.

**Resolve Fallout.Common version from NuGet, scoped to the tool's own major** — `ResolveFalloutVersionStep` now queries NuGet's flat-container index for the latest non-prerelease `Fallout.Common` release matching the running tool's own major (calendar year), instead of only reading the tool's own `AssemblyInformationalVersion`. Falls back to the tool's own version when NuGet can't be reached or has no matching-major stable release yet.

**Warn when the build project targets an older TFM than .NET 10** — `VerifyBuildTargetFrameworkStep` reads `_build.csproj`'s `TargetFramework(s)` and warns for any moniker older than `net10.0`.

**Ask for confirmation before the migration writes any files** — `ConfirmMigrationStep` prompts via `AnsiConsole.Confirm` immediately before the first file-mutating step. Declining sets `Summary.Cancelled` and stops the pipeline. Skipped for `--dry-run` and when stdin is redirected.

**Fold CsprojRewriter into RewriteCsprojsStep** — `CsprojRewriter.Rewrite` had a single caller; merged directly into the step.

**Bump build project to net10.0 and global.json SDK to 10.0.100** — `BumpDotNetVersionStep` rewrites `_build.csproj`'s `TargetFramework` and `global.json`'s `sdk.version` when either is behind the minimum, leaving already-current or newer values untouched. Build projects don't multi-target, so only the singular `TargetFramework` element is matched.

**Make IMigrationStep and the migration pipeline asynchronous** — `IMigrationStep.Execute` is now `Task ExecuteAsync`; replaces the blocking `GetAsync(...).GetAwaiter().GetResult()` in `ResolveFalloutVersionStep` with a proper `await`.

**Deduplicate the TFM-minimum check shared by two migration steps** — extract the modern-moniker regex and minimum-check logic duplicated between `BumpDotNetVersionStep` and `VerifyBuildTargetFrameworkStep` into a shared `TargetFrameworkMonikers` helper.

**Reorder ResolveFalloutVersionStep by invocation order and tag NuGet requests** — reorder private methods to match invocation order; add a `User-Agent` header to the NuGet flat-container `HttpClient` so requests aren't sent as anonymous script traffic.

## Combined effect
`fallout-migrate` now pins migrated references to a real, published `Fallout.Common` release, flags/fixes an outdated build-project TFM, confirms before mutating files, and runs its pipeline asynchronously.


src/Fallout.Migrate/Steps/BumpDotNetVersionStep.cs
```diff
@@ -0,0 +1,90 @@
...
+{
+    /// <summary>
+    /// The .NET target framework moniker to bump <c>_build.csproj</c> to.
+    /// </summary>
+    private const string TargetFramework = "net10.0";
+
+    /// <summary>
+    /// The .NET SDK version to pin <c>global.json</c> to.
+    /// </summary>
+    private const string SdkVersion = "10.0.100";
```
IT-VBFK, line 24: I don't think users who migrate will expect something like this...

Suppose I have my version at 10.0.300 (because I want it to) the migration step would downgrade the SDK..

But the other way round is probably valid: I have set it to 9.0.x and bump it to 10.0.x

So.. I think this could be either an option and let the user decide if downgrade the global SDK or just upgrade but never downgrade

src/Fallout.Migrate/Steps/ConfirmMigrationStep.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        if (Console.IsInputRedirected)
+        {
+            // Non-interactive (CI, piped input, automated tests) — nothing to prompt against, so
+            // proceed as if confirmed.
+            return;
+        }
+
+        bool proceed = AnsiConsole.Confirm(
+            $"This will modify files under [blue]{context.RootDirectory}[/]. Continue?");
```
IT-VBFK, line 34: Since the migration does more than "just" switching from NUKE to Fallout, maybe we should give the use an exact list of what changes.

I speek for this "additional", possibly unintended changes like sdk pin, ...

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
```diff
@@ -1,43 +1,150 @@
...
+    /// version matches <paramref name="major"/>.
+    /// </summary>
+    /// <param name="major">The major version (calendar year) to restrict the lookup to.</param>
+    /// <returns>The latest published stable version within <paramref name="major"/>, or <c>null</c>
+    /// if NuGet couldn't be reached or has no matching-major stable version.</returns>
+    private static string ResolveFromNuGet(int major)
+    {
+        try
+        {
+            using HttpResponseMessage response = httpClient.GetAsync(FlatContainerIndex).GetAwaiter().GetResult();
```
IT-VBFK: Can this be "real async" here?

src/Fallout.Migrate/Steps/BumpDotNetVersionStep.cs
dennisdoomen: Agreed. I'll make sure the code base matches the minimum version.

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
dennisdoomen: Yeah, was thinking about that as well.

src/Fallout.Migrate/Steps/ConfirmMigrationStep.cs
dennisdoomen: No need. That's why they have source control.

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
IT-VBFK: nice one

src/Fallout.Migrate/Common/TargetFrameworkMonikers.cs
ChrisonSimtian, line 1: probably something that could move a bit closer to the core instead of sitting in this satellite. Looks useful :-)

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
```diff
@@ -1,43 +1,83 @@
...
 /// Must run first in <see cref="Migration.steps"/>: <see cref="Steps.RewriteCsprojsStep"/> reads the
 /// resolved version to pin rewritten package references.
 /// </summary>
 internal sealed class ResolveFalloutVersionStep : IMigrationStep
 {
-    /// <summary>The version to fall back to when the assembly carries no build-metadata suffix.</summary>
+    /// <summary>
+    /// The version to fall back to when the assembly carries no build-metadata suffix.
+    /// </summary>
     private const string Fallback = "10.3.49";
```
ChrisonSimtian, line 35: how do we remember to bump this with a new release?
Should this be more central? collection of constants somewhere grouped together as a one-stop-shop maybe?

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
```diff
@@ -1,43 +1,83 @@
...
+    /// NuGet's flat-container index for <see cref="PackageId"/>: a JSON document listing every
+    /// published version (stable and prerelease), oldest first.
+    /// </summary>
+    private static readonly Uri FlatContainerIndex = new($"https://api.nuget.org/v3-flatcontainer/{PackageId}/index.json");
+
+    private static readonly HttpClient httpClient = CreateHttpClient();
+
     /// <inheritdoc />
-    public void Execute(MigrationContext context, Summary summary)
+    public async Task ExecuteAsync(MigrationContext context, Summary summary)
```
ChrisonSimtian, line 51: no CancellationToken?

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
```diff
@@ -1,43 +1,83 @@
...
     {
-        context.FalloutVersion = Resolve();
+        string localVersion = ResolveFromAssembly();
+        int localMajor = ExtractMajor(localVersion);
+
+        string nuGetVersion = await ResolveFromNuGetAsync(localMajor);
```
ChrisonSimtian, line 56: no CancellationToken?

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
```diff
@@ -54,4 +94,86 @@ private static string Resolve()
...
+    }
+
+    /// <summary>
+    /// Queries NuGet for the latest non-prerelease version of <see cref="PackageId"/> whose major
+    /// version matches <paramref name="major"/>.
+    /// </summary>
+    /// <param name="major">The major version (calendar year) to restrict the lookup to.</param>
+    /// <returns>The latest published stable version within <paramref name="major"/>, or <c>null</c>
+    /// if NuGet couldn't be reached or has no matching-major stable version.</returns>
+    private static async Task<string> ResolveFromNuGetAsync(int major)
```
ChrisonSimtian, line 119: no CancellationToken?

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
```diff
@@ -54,4 +94,86 @@ private static string Resolve()
...
+    {
+        try
+        {
+            using HttpResponseMessage response = await httpClient.GetAsync(FlatContainerIndex);
```
ChrisonSimtian: pretty sure this one lets us pass through a token, mind having a look at that and passing it all the way through if so?

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
```diff
@@ -54,4 +94,86 @@ private static string Resolve()
...
+        }
+    }
+
+    /// <summary>
+    /// Creates the <see cref="HttpClient"/> used for the NuGet flat-container lookup, identifying
+    /// this tool via a <c>User-Agent</c> header so NuGet.org doesn't treat the request as anonymous
+    /// script traffic.
+    /// </summary>
+    /// <returns>A configured <see cref="HttpClient"/>.</returns>
+    private static HttpClient CreateHttpClient()
```
ChrisonSimtian, line 172: nit: can we make this a factory instead?

ChrisonSimtian (commented):
some small things :-)

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
ITaluone: I think there is no urgent need to bump this on _every_ release. Since this is a fallback, only used when this very assembly not carrying a version (and only the major part is used to lookup from nuget), I think once on major releases should suffice.

ITaluone (approved):

src/Fallout.Migrate/Steps/ScriptRewriter.cs
ITaluone, line 11: One little thing to propose here:

In NUKE's bootstrapper files (`build.sh` and `build.ps1`) there usually is a section like

```ps1
if (Test-Path env:NUKE_ENTERPRISE_TOKEN) {
    & $env:DOTNET_EXE nuget remove source "nuke-enterprise" > $null
    & $env:DOTNET_EXE nuget add source "https://f.feedz.io/nuke/enterprise/nuget" --name "nuke-enterprise" --username "PAT" --password $env:NUKE_ENTERPRISE_TOKEN > $null
}
```

or 
```bash
if [[ ! -z ${NUKE_ENTERPRISE_TOKEN+x} && "$NUKE_ENTERPRISE_TOKEN" != "" ]]; then
    "$DOTNET_EXE" nuget remove source "nuke-enterprise" &>/dev/null || true
    "$DOTNET_EXE" nuget add source "https://f.feedz.io/nuke/enterprise/nuget" --name "nuke-enterprise" --username "PAT" --password "$NUKE_ENTERPRISE_TOKEN" --store-password-in-clear-text &>/dev/null || true
fi
```

which I think was already obsolete in NUKE at the end 🤔

Should we also consider removing these legacy lines too? Just to prevent to accumulate legacy code parts

ChrisonSimtian: I'd raise that as a separate tidy up PR, not the end of the world but agreed, this can go.
I dont think we need this and I dont think should we ever decide to monetize on this, that this is the way I'd envision it to go ;-)

ChrisonSimtian (approved):
i think i forgot to greentick this last time ... sorry bit forgetful right now :-)

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
dennisdoomen: @ITaluone is right. And when we're shipping a major release of Fallout, we'll also ship a major release of Fallout.Migrate, which will have the v11 moniker. So, in reality, we'll never have to bump the version.

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
dennisdoomen: Nah. It's a CLI that completes in seconds, so I didn't bother. Do you feel it's important?

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
dennisdoomen: What token do you mean? The tool doesn't offer that option.

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
dennisdoomen: A factory? For what?

src/Fallout.Migrate/Steps/ScriptRewriter.cs
dennisdoomen: See #522

IT-VBFK: Or #518 / #519

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
IT-VBFK: CancellationToken for sure :)

But I want to see this guy who can cancel the ongoing migration 😂 (the whole migration takes less than a second)

dennisdoomen:
@ChrisonSimtian As you already approved the PR, I assume your comments were suggestions. If you feel they need addressing, let me know. More than fine creating another PR.

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
dennisdoomen: Ah, I was somehow thinking of a NuGet token of some sort. 🤦‍♂️

PR #511, opened 2026-07-19
Fix crash when git branch config has a duplicate key

dennisdoomen:
Fixes a crash in `GitRepository.GetRemoteNameAndBranch` when the git branch config section contains a duplicate key.

## Problem
VS Code's Git extension appends a `vscode-merge-base` entry to `[branch "..."]` instead of updating it in place, sometimes leaving it duplicated. Parsing that section built a `Dictionary` with `ToDictionary`, which throws:

```
System.ArgumentException: An item with the same key has already been added. Key: vscode-merge-base
```

This breaks `GitRepository.FromLocalDirectory`, and therefore `[GitVersion]` parameter injection, for any consumer whose `.git/config` has this duplicate.

## Fix
Group config lines by key and take the last value (matches git's own last-value-wins semantics) instead of `ToDictionary`. Also skip lines without `=` instead of crashing with `IndexOutOfRangeException`.

## Testing
Added a regression test that reproduces the exact reported exception against a real temp git repo with a duplicated `vscode-merge-base` line, and confirmed it fails without the fix and passes with it. Ran the full `Fallout.Build.Tests` suite (only the pre-existing, unrelated `FromDirectoryTest` worktree limitation fails, confirmed to fail identically on `main`).

### DIFF
--- a/tests/Fallout.Build.Specs/GitRepositorySpecs.cs
+++ b/tests/Fallout.Build.Specs/GitRepositorySpecs.cs
@@ -74,4 +74,71 @@ public class GitRepositorySpecs
                 Directory.Delete(tempDir, recursive: true);
         }
     }
+
+    /// <summary>
+    /// VS Code's Git extension is known to append a duplicate `vscode-merge-base` line to the
+    /// `[branch "name"]` section instead of updating it in place, which used to crash
+    /// GetRemoteNameAndBranch's ToDictionary call.
+    /// </summary>
+    [Fact]
+    public void Handles_duplicate_config_keys_caused_by_vs_code()
+    {
+        // Arrange
+        var directory = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString("N"));
+        Directory.CreateDirectory(directory);
+
+        try
+        {
+            RunGit(directory, "init --quiet --initial-branch=main");
+            RunGit(directory, "config user.email test@example.com");
+            RunGit(directory, "config user.name Test");
+            File.WriteAllText(Path.Combine(directory, "file.txt"), "content");
+            RunGit(directory, "add file.txt");
+            RunGit(directory, "commit --quiet -m initial");
+            // Deliberately not a real Fallout-build/Fallout URL: some CI environments rewrite
+            // that org's URLs (e.g. to a canary mirror) via a global git `insteadOf` config,
+            // which would silently change the identifier this test asserts on.
+            RunGit(directory, "remote add origin https://github.com/octocat/hello-world.git");
+            RunGit(directory, "config branch.main.remote origin");
+            RunGit(directory, "config branch.main.merge refs/heads/main");
+
+            var configPath = Path.Combine(directory, ".git", "config");
+            File.AppendAllText(
+                configPath,
+                "\tvscode-merge-base = origin/main" + Environment.NewLine +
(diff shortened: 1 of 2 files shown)


tests/Fallout.Build.Specs/GitRepositorySpecs.cs
```diff
@@ -74,4 +74,68 @@ public void Fails_for_non_git_directory()
...
+
+        try
+        {
+            RunGit(directory, "init --quiet --initial-branch=main");
+            RunGit(directory, "config user.email test@example.com");
+            RunGit(directory, "config user.name Test");
+            File.WriteAllText(Path.Combine(directory, "file.txt"), "content");
+            RunGit(directory, "add file.txt");
+            RunGit(directory, "commit --quiet -m initial");
+            RunGit(directory, "remote add origin https://github.com/nuke-build/nuke.git");
```
IT-VBFK:
```suggestion
            RunGit(directory, "remote add origin https://github.com/Fallout-build/Fallout.git");
```

Irrelevant what remote repo is here, but just to align

tests/Fallout.Build.Specs/GitRepositorySpecs.cs
```diff
@@ -74,4 +74,68 @@ public void Fails_for_non_git_directory()
...
+                configPath,
+                "\tvscode-merge-base = origin/main" + Environment.NewLine +
+                "\tvscode-merge-base = origin/main" + Environment.NewLine);
+
+            // Act
+            var repository = GitRepository.FromLocalDirectory(directory);
+
+            // Assert
+            repository.Branch.Should().Be("main");
+            repository.Identifier.Should().Be("nuke-build/nuke");
```
IT-VBFK:
```suggestion
            repository.Identifier.Should().Be("Fallout-build/Fallout");
```

tests/Fallout.Build.Specs/GitRepositorySpecs.cs
dennisdoomen: Applied, thanks.

tests/Fallout.Build.Specs/GitRepositorySpecs.cs
dennisdoomen: Applied, thanks.

tests/Fallout.Build.Specs/GitRepositorySpecs.cs
```diff
@@ -74,4 +74,68 @@ public void Fails_for_non_git_directory()
...
+
+        try
+        {
+            RunGit(directory, "init --quiet --initial-branch=main");
+            RunGit(directory, "config user.email test@example.com");
+            RunGit(directory, "config user.name Test");
+            File.WriteAllText(Path.Combine(directory, "file.txt"), "content");
+            RunGit(directory, "add file.txt");
+            RunGit(directory, "commit --quiet -m initial");
+            RunGit(directory, "remote add origin https://github.com/Fallout-build/Fallout.git");
```
ChrisonSimtian:
```suggestion
            RunGit(directory, "remote add origin https://github.com/Fallout-build/Fallout.Canary.git");
```

I'd do the test against our canary, instead of the main repo. Just to set a good example and not tempt people / ai to accidentally run tests against the live repo :-)

ChrisonSimtian (approved):
lgtm, see suggestion in your test

PR #512, opened 2026-07-19
Fail CI when generated tool wrappers are out of sync with their .json specs

dennisdoomen:
**Problem**

`GenerateTools` regenerates the `.Generated.cs` tool wrappers from their `.json` specs, but it's a manual target with no CI enforcement. A contributor can edit a tool's `.json` spec without regenerating its wrapper, merge silently, and ship stale wrapper code in the next release.

**Outcome**

Adds a `VerifyGeneratedTools` target that regenerates the tool wrappers from a known-clean checkout and asserts the working copy stays clean afterward, failing the build if regeneration produces a diff. Wires it into the `ubuntu-latest` required PR gate ahead of `Test` and `Pack`, so drift between a `.json` spec and its generated wrapper is caught before merge instead of shipping in the next release.

Includes the regenerated `.fallout/build.schema.json` and `.github/workflows/ubuntu-latest.yml` (both generated from `build/Build.CI.GitHubActions.cs` / the target list — not hand-edited).


IT-VBFK:
So: you want the contributors to commit the generated files? Then we have to fix two things:
- The Contributing guidelines
- The comment generated in every wrapper's header, which contains the remote repo of the current executing user. This is the crucial one, because otherwise it will fail _everytime_

dennisdoomen:
> So: you want the contributors to commit the generated files? Then we have to fix two things:

The docs are wrong. Those generated files were never regenerated unless you ran the `GenerateTools` target. I'll update the docs as well, but wasn't ready yet, hence a _draft_ PR ;-)

IT-VBFK:
Would be quite fancy, if the CI's generated tool wrappers get committed automagically. Like: CI re-generates tools -> CI commits (the new ones) -> pushes to the PR.

So, it would always fit and nothing to forget about.

dennisdoomen:
> Would be quite fancy, if the CI's generated tool wrappers get committed automagically. Like: CI re-generates tools -> CI commits (the new ones) -> pushes to the PR.
> 
> So, it would always fit and nothing to forget about.

I'd prefer a manual check that the generated file makes sense.

IT-VBFK:
> The docs are wrong. Those generated files were never regenerated unless you ran the GenerateTools target. 

Maybe this was a manual step done by Matthias before a release.. But running by contributors and check in CI/CD is more convenient :)

IT-VBFK:
Missing the update here: https://github.com/Fallout-build/Fallout/blob/c99b2384e04ccaad7fe90a02dea920616bbeb393/CONTRIBUTING.md?plain=1#L56

meaning: align to the agents/conventions.md

`- Commit the regenerated `.cs` output alongside the `.json` spec — `VerifyGeneratedTools` fails CI if they drift.`

ChrisonSimtian (approved):

PR #515, opened 2026-07-20
Remove unused `buildScript` and `args` parameters from commands

ITaluone:
Fixing nasty build errors due to the .editorconfig rule `UnusedParameter.Local`'s severity `error`

This only applies to the private `Execute` methods; the `IFalloutCommand` contract is not touched.

When needed, we can re-add them..


ChrisonSimtian:
ah yeah I noticed those, they are everywhere ... well done

ChrisonSimtian (approved):

dennisdoomen (approved):

Issue #203, opened 2026-05-26
Deprecate per-repo bootstrappers; make Fallout.GlobalTool the canonical entry point


ChrisonSimtian:
## Why

Today every Fallout-using repo carries \`build.cmd\` (7 lines), \`build.sh\` (~70 lines), and \`build.ps1\` (~70 lines). They do two jobs:

1. Provision the .NET SDK locally (\`dotnet-install.sh/ps1\` → \`.fallout/temp/\`) when it's missing.
2. Compile + run \`build/_build.csproj\` via \`dotnet build\` + \`dotnet run --project --no-build\`.

The \`fallout\` global tool exists but, for target execution, just shells out to \`bash ./build.sh\` — it's a launcher, not a runner. We ship a .NET build tool whose canonical entry point is a shell script. That's the thing we want to fix.

Goal: one logical entry point (\`fallout\`), one C# code path for target execution, no \`.cmd\`, and thin shell shims only for the "I just cloned the repo without \`dotnet\` installed yet" case.

## Approach

Three PRs, sequenced. The split keeps PR-A as a pure additive change (the in-tool runner ships dormant; CI doesn't depend on it yet) so it can be released independently. Downstream consumers who upgrade \`Fallout.GlobalTool\` pick up the new runner; their old fat shims still work; they can adopt thin shims on their own schedule by re-running \`fallout :setup\`.

### PR-A — Add in-tool build runner (#201)

**Status: open, code+tests done locally, waiting on GitHub Actions to recover (see "Known issues" below).**

Adds a code path so \`fallout <target>\` discovers and runs the build project directly, instead of spawning \`./build.sh\`/\`./build.ps1\`.

- **New** \`src/Fallout.GlobalTool/BuildProjectResolver.cs\` — reads optional \`BuildProjectFile\` field from \`.fallout/parameters.json\` if set; otherwise convention \`build/_build.csproj\`. 6 unit tests cover the resolution cases.
[...]

PR #517, opened 2026-07-20
Use `dotnet fallout` instead of ./build.cmd in AppVeyor and Azure Pipelines steps

### LINKED ISSUE #516: GitHub Actions run step hardcodes ./build.cmd — breaks non-Windows runners (fix: pwsh build.ps1)
## Summary

`GitHubActionsRunStep.Write()` emits `run: ./{BuildCmdPath} <targets>` for every job regardless of the runner image, and `ConfigurationAttributeBase.BuildCmdPath` only globs `build.cmd` (never `build.sh`/`build.ps1`). So any workflow generated for an `Ubuntu*`/`MacOs*` image gets `./build.cmd`, which can't execute — only `windows-*` runners work.

## Repro

```csharp
[GitHubActions("ci", GitHubActionsImage.UbuntuLatest, InvokedTargets = [nameof(Foo)])]
```
generates:
```yaml
runs-on: ubuntu-latest
...
  run: ./build.cmd Foo      # .cmd on ubuntu → fails
```

## Cause (10.3.49)

- `Fallout.Common.CI.GitHubActions.Configuration.GitHubActionsRunStep.Write` → `"  run: ./" + BuildCmdPath + " " + JoinSpace(InvokedTargets)`
- `Fallout.Common.CI.ConfigurationAttributeBase.BuildCmdPath` → globs `build.cmd`, `*/build.cmd` only

No branch on `GitHubActionsImage`. Confirmed across both assemblies: **zero** references to `build.sh` or `build.ps1`; `build.cmd` is the only build-script string (2 glob literals in `Fallout.Build`). The `build.sh`/`build.ps1` bootstrap files ship, but the generator only ever knows `build.cmd`.

## Fix — pwsh route (decided)

`pwsh` is on every GitHub-hosted runner (win/linux/mac) and `build.ps1` already takes targets positionally (`ValueFromRemainingArguments`), so invoke pwsh and drop per-OS script selection entirely:

1. `ConfigurationAttributeBase` — add a pwsh path mirroring `BuildCmdPath`:
   ```csharp
   protected virtual string BuildScriptPath =>
       Build.RootDirectory.GlobFiles("build.ps1", "*/build.ps1")
           .Select(x => Build.RootDirectory.GetUnixRelativePathTo(x))
           .FirstOrDefault().NotNull();
   ```
2. `GitHubActionsRunStep.Write` — invoke pwsh:
   ```csharp
   writer.WriteLine("  run: pwsh -File ./" + BuildScriptPath + " " + JoinSpace(InvokedTargets));
   ```
   (thread `BuildScriptPath` through the attribute's `GetSteps` where it currently passes `BuildCmdPath`.)

`pwsh -File ./build.ps1 <targets>` runs regardless of the runner's default shell — no `shell:` key, no image branch. Same swap applies to the other CI run-step writers if we want them pwsh too.

## Workaround

Pin `GitHubActionsImage.WindowsLatest` (only image where the generated `./build.cmd` runs).

---
*Filed by AI (Claude Opus 4.8 (1M context)) on Chris's behalf.*

ChrisonSimtian:
Fixes #516 for the providers where the bug still lives on `main`.

## Problem
AppVeyor and Azure Pipelines emit a `./build.cmd` invocation that only runs on Windows agents:
- **AppVeyor** emitted `- sh: ./build.cmd …` — a Unix shell running a `.cmd`.
- **Azure** ran `script: './build.cmd …'` in `CmdLine@2` regardless of agent OS.

Both break on Ubuntu/macOS agents. (GitHub Actions — #516's original repro — was already fixed on the 2026 line via `dotnet fallout` in #203/#204, so it needs no change here.)

## Fix
Same shape as the GitHub Actions fix: invoke the cross-platform `fallout` local tool.
- AppVeyor: `dotnet tool restore` + `dotnet fallout <targets>` (emitted under both `- cmd:` and `- sh:` so each image runs the matching pair; command is identical).
- Azure: `script: 'dotnet tool restore && dotnet fallout <targets>'` (`&&` works in cmd and bash; `CmdLine@2` uses the agent's default shell).

No OS branching, no build-script selection.

## Out of scope
- TeamCity and SpaceAutomation already select `.sh` on non-Windows — not broken, left unchanged.
- `BuildCmdPath` stays on the config types (public API) but is no longer consumed by these two writers.

## Verification
Regenerated Verify snapshots; `ConfigurationGenerationSpecs` 18/18 green.

### DIFF
--- a/src/Fallout.Common/CI/AppVeyor/Configuration/AppVeyorConfiguration.cs
+++ b/src/Fallout.Common/CI/AppVeyor/Configuration/AppVeyorConfiguration.cs
@@ -110,8 +110,13 @@ public class AppVeyorConfiguration : ConfigurationEntity
 
         using (writer.WriteBlock("build_script:"))
         {
-            writer.WriteLine($@"- cmd: .\{BuildCmdPath} {InvokedTargets.JoinSpace()}");
-            writer.WriteLine($@"- sh: ./{BuildCmdPath} {InvokedTargets.JoinSpace()}");
+            // Invoke the `fallout` local tool (cross-platform) rather than the OS-specific build script.
+            // AppVeyor runs `- cmd:` lines only on Windows images and `- sh:` lines only on the rest, so
+            // emit both; the command itself is identical. See GitHub Actions (dotnet fallout) and issue #516.
+            writer.WriteLine("- cmd: dotnet tool restore");
+            writer.WriteLine($"- cmd: dotnet fallout {InvokedTargets.JoinSpace()}");
+            writer.WriteLine("- sh: dotnet tool restore");
+            writer.WriteLine($"- sh: dotnet fallout {InvokedTargets.JoinSpace()}");
         }
 
         if (Cache.Length > 0)
--- a/src/Fallout.Common/CI/AzurePipelines/Configuration/AzurePipelinesCmdStep.cs
+++ b/src/Fallout.Common/CI/AzurePipelines/Configuration/AzurePipelinesCmdStep.cs
@@ -25,7 +25,10 @@ public class AzurePipelinesCmdStep : AzurePipelinesStep
 
             using (writer.WriteBlock("inputs:"))
             {
-                writer.WriteLine($"script: './{BuildCmdPath} {arguments}'");
+                // Invoke the `fallout` local tool (cross-platform) rather than the OS-specific build script.
+                // CmdLine@2 runs in the agent's default shell (cmd on Windows, bash elsewhere); `&&` works in
+                // both. See GitHub Actions (dotnet fallout) and issue #516.
+                writer.WriteLine($"script: 'dotnet tool restore && dotnet fallout {arguments}'");
             }
 
             if (Imports.Count > 0)
--- a/tests/Fallout.Common.Specs/CI/ConfigurationGenerationSpecs.Test_testName=null_attribute=AppVeyorAttribute.verified.txt
+++ b/tests/Fallout.Common.Specs/CI/ConfigurationGenerationSpecs.Test_testName=null_attribute=AppVeyorAttribute.verified.txt
@@ -31,8 +31,10 @@ install:
   - git submodule update --init --recursive
 
 build_script:
-  - cmd: .\build.cmd Test
-  - sh: ./build.cmd Test
+  - cmd: dotnet tool restore
+  - cmd: dotnet fallout Test
+  - sh: dotnet tool restore
+  - sh: dotnet fallout Test
 
 artifacts:
   - path: 'src/*/obj/**'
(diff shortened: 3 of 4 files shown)


ITaluone:
Ahmm.. `build.cmd` can execute on Unix, but the requirements have to meet: `chmod +x` (which is probably not set for `build.cmd`)

The reasons why this should work are:
- extensions are meaningless on unix
- the first part targets unix systems (`:` is a comment in windows cmd, but a command on unix -> `:` = the "null command" making this a valid syntax (on windows -> line comment, on unix command)

I think the reason why this doesn't work here: there is no build.cmd file anymore :)

Ou.. and: the `:` operator only works in bash not in pure `/bin/sh`, maybe this was an error as well?

(script path should be quoted: `"${SCRIPT_DIR}/build.sh" "$@"`)

Edit: I just tested this against a local project of mine (unix/docker executor).. works fine for me...

ChrisonSimtian:
> Ahmm.. `build.cmd` can execute on Unix, but the requirements have to meet: `chmod +x` (which is probably not set for `build.cmd`)
> 
> The reasons why this should work are:
> 
> * extensions are meaningless on unix
> * the first part targets unix systems (`:` is a comment in windows cmd, but a command on unix -> `:` = the "null command" making this a valid syntax (on windows -> line comment, on unix command)
> 
> I think the reason why this doesn't work here: there is no build.cmd file anymore :)
> 
> Ou.. and: the `:` operator only works in bash not in pure `/bin/sh`, maybe this was an error as well?
> 
> (script path should be quoted: `"${SCRIPT_DIR}/build.sh" "$@"`)
> 
> Edit: I just tested this against a local project of mine (unix/docker executor).. works fine for me...

typical claude PR description ... the whole point of this PR was to go and unify the build entry to use dotnet.
Here is my narrative for changing it:
you wanna build dotnet, therefore a requirement is dotnet is installed on the build agent. Hence we can use dotnet to run fallout, why bother with OS agnostic scripting languages as a bootstrapper?

Claude just tripped over it because we got windows AND linux runners at work and it tried to spin up a linux runner but found the windows one hardcoded, hence made the assumption its because of the CMD (which in return triggered me to say "hey holdup, I changed this already, havent I?")

So all good, just ignore the blurb at the beginning :-)
PR also shouldve been raised in draft, which obviously also went hayway... Must be either my cold making me prompt silly or claude simply being made dumber now that fable is out :D

ITaluone:
Ok... got it :)

But this will change the behavior for consumers, right? IMHO we should not change the "build agent agnostic" behavior, because it has a real use-case by installing and setting up all which is needed for the pipeline to run.

ChrisonSimtian:
hence the draft status. Github already does that, my reasoning here is if GH works that way why have the other platforms work differently?
But I also got other plans for CI platforms, so lets see where that goes and what comes in first

PR #519, opened 2026-07-20
Add a cleanup step for bootstrapper files to remove stale `NUKE_ENTERPRISE_TOKEN` handling

### LINKED ISSUE #518: [Migration] - Remove legacy code parts: Stale NUKE_ENTERPRISE_TOKEN handling in bootstrapper files
### Description

Previous version of NUKE added a check for `NUKE_ENTERPRISE_TOKEN` and adds a special nuget feed in `build.sh` and `build.ps1`.

I propose to let the migration get rid of this because:
- At time of writing there is no plan go commercial
- If this env var is set for some users, they will not have an advantage, since this feed at least is stale, at worst non existent anymore

### Usage Example

Just with the common `fallout-migrate` command doing its work

### Alternative

Leave as is, let the user remove this when they step over. But why don't give the users a service by removing legacy code parts?

Yes

### LINKED ISSUE #522: Strip legacy NUKE_ENTERPRISE_TOKEN blocks from bootstrap scripts during migration
### Problem
`ScriptRewriter` (used by `RewriteBootstrapScriptsStep`) rewrites `build.cmd`/`build.ps1`/`build.sh` but leaves behind legacy `NUKE_ENTERPRISE_TOKEN` feed-registration blocks (the `nuke-enterprise` NuGet source add/remove lines), which were already dead in NUKE. Migrated repos carry this dead code forward.

### Outcome
`fallout-migrate` strips these obsolete `NUKE_ENTERPRISE_TOKEN` blocks from bootstrap scripts during rewrite, instead of preserving them.

### Acceptance criteria
- [ ] `ScriptRewriter` removes the `NUKE_ENTERPRISE_TOKEN` conditional block (both `.ps1` and `.sh` variants) when present
- [ ] Existing bootstrap scripts without this block are unaffected

### Notes
From [#509 (review comment)](https://github.com/Fallout-build/Fallout/pull/509#discussion_r3612693093).

ITaluone:
Closes: #518 

Add a small cleanup step, which removes stale env var check for the `NUKE_ENTERPRISE_TOKEN` variable, which in return adds a special nuget source (which at best is stale and at worst non-existent).

- Added direct cleanup tests
- Additional MigrationsIntegrationTests

### DIFF
--- a/src/Fallout.Migrate/Steps/CleanupBootstrapScriptsStep.cs
+++ b/src/Fallout.Migrate/Steps/CleanupBootstrapScriptsStep.cs
@@ -1,0 +1,71 @@
+using System;
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using Fallout.Common.IO;
+using Fallout.Migrate.Common;
+
+namespace Fallout.Migrate.Steps;
+
+/// <summary>
+/// This migration step cleans up bootstrap scripts by removing unnecessary code.
+/// The unnecessary code includes checks for a specific environment variable and related logic.
+/// </summary>
+internal partial class CleanupBootstrapScriptsStep : IMigrationStep
+{
+    public Task ExecuteAsync(MigrationContext context, Summary summary)
+    {
+        foreach (var file in new[]
+                 {
+                     "build.sh",
+                     "build.ps1"
+                 })
+        {
+            var path = context.RootDirectory / file;
+            if (path.FileExists())
+            {
+                MigrationFileOperations.ApplyRewrite(context, path, Cleanup, summary);
+            }
+        }
+
+        return Task.CompletedTask;
+    }
+
+    private static RewriteResult Cleanup(string content)
+    {
+        var envVarToCheck = "NUKE_ENTERPRISE_TOKEN";
+        if (!content.Contains(envVarToCheck))
+        {
+            return new(content, 0);
+        }
+
+        // This is just to preserve the current line ending the user has for this files
+        string newline =
+            GetLineFeedRegex().Match(content).Value is { Length: > 0 } value
+                ? value
+                : Environment.NewLine;
+
+        var lines = GetLineFeedRegex().Split(content).ToList();
+
+        // here, we get the index of the line that contains the environment variable check
+        var indexOfEnterpriseEnvVarCheck = lines.FindIndex(line => line.Contains(envVarToCheck));
+        
+        // here, we get the index of the line that ends the if block
+        // which is fi on bash and a simple "}" in powershell
+        var endOfIfBlock = lines.FindIndex(indexOfEnterpriseEnvVarCheck,
+            line => line.Trim() == "}" || line.Trim() == "fi");
+
+        // this is "just" to remove the empty line after the if block
+        if (lines.Count > endOfIfBlock + 1 && lines[endOfIfBlock + 1].Trim() == "")
+        {
+            endOfIfBlock++;
+        }
+
+        lines.RemoveRange(indexOfEnterpriseEnvVarCheck, endOfIfBlock - indexOfEnterpriseEnvVarCheck + 1);
+
+        return new(string.Join(newline, lines), 1);
+    }
+
+    [GeneratedRegex(@"\r\n|\n|\r")]
+    private static partial Regex GetLineFeedRegex();
+}
--- a/docs/migration/from-nuke.md
+++ b/docs/migration/from-nuke.md
@@ -66,8 +66,9 @@ If your build had previously worked against NUKE, it should now work against Fal
 | Fully-qualified type refs | `Nuke.Common.AbsolutePath` | `Fallout.Common.AbsolutePath` |
 | Base class name | `: NukeBuild` | `: FalloutBuild` |
 | Base interface name | `: INukeBuild` | `: IFalloutBuild` |
-| MSBuild properties in `_build.csproj` | `<NukeRootDirectory>` | `<FalloutRootDirectory>` |
-| Bootstrap scripts | `dotnet nuke`, `.nuke/temp` | `dotnet fallout`, `.fallout/temp` |
+| MSBuild properties in `_build.csproj` | `<NukeRootDirectory>`, `<NukeTelemetryVersion>` | `<FalloutRootDirectory>`, `<FalloutTelemetryVersion>` |
+| Bootstrap scripts | `dotnet nuke`, `NUKE_TELEMETRY_OPTOUT`, `.nuke/temp` | `dotnet fallout`, `FALLOUT_TELEMETRY_OPTOUT`, `.fallout/temp` |
+| Removes stale Nuke Enterprise NuGet source | The whole env var check for `NUKE_ENTERPRISE_TOKEN` and its handling | The bootstrapper files (`build.sh` and `build.ps1`) without these `if`'s |
 | Telemetry knobs (removed) | `<NukeTelemetryVersion>`, `NUKE_TELEMETRY_OPTOUT` | *dropped* — Fallout has no telemetry ([ADR-0010](../adr/0010-no-telemetry-collection.md)); `fallout migrate` strips them |
 | Config directory | `.nuke/` | `.fallout/` (contents preserved) |
 
--- a/src/Fallout.Migrate/Migration.cs
+++ b/src/Fallout.Migrate/Migration.cs
@@ -30,6 +30,7 @@ internal sealed class Migration(AbsolutePath rootDirectory, bool dryRun, TextWri
         new BumpDotNetVersionStep(),
         new RewriteCsFilesStep(),
         new RewriteBootstrapScriptsStep(),
+        new CleanupBootstrapScriptsStep(),
         new RenameNukeDirectoryStep()
     ];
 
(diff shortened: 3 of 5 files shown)


IT-VBFK:
@dennisdoomen Should this go into the already existing `RewriteBootstrapScriptsStep.cs` or like this as dedicated step?

dennisdoomen:
> Should this go into the already existing `RewriteBootstrapScriptsStep.cs` or like this as dedicated step?

I would definitely make that a dedicated step.

IT-VBFK:
Cool :)

ITaluone:
Should now fit definitely :)

src/Fallout.Migrate/Steps/CleanupBootstrapScriptsStep.cs
```diff
@@ -0,0 +1,62 @@
...
+            if (path.FileExists())
+            {
+                MigrationFileOperations.ApplyRewrite(context, path, Cleanup, summary);
+            }
+        }
+
+        return Task.CompletedTask;
+    }
+
+    internal static RewriteResult Cleanup(string content)
```
dennisdoomen: 🔧 We don't want to test the implementation details of a step. Instead, we use the public API of this class. 

🌱 (planting a seed) In principle testing the step directly is wrong as well. We should really be using the `Migration` class as the subject.

src/Fallout.Migrate/Steps/CleanupBootstrapScriptsStep.cs
```diff
@@ -0,0 +1,62 @@
+using System;
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using Fallout.Common.IO;
+using Fallout.Migrate.Common;
+
+namespace Fallout.Migrate.Steps;
+
+internal partial class CleanupBootstrapScriptsStep : IMigrationStep
```
dennisdoomen, line 14: 🔧 Some documentation on what this step does is not optional

src/Fallout.Migrate/Steps/CleanupBootstrapScriptsStep.cs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    internal static RewriteResult Cleanup(string content)
+    {
+        if (!nukeEnterprisePattern.IsMatch(content))
+        {
+            return new(content, 0);
+        }
+
+        string newline =
```
dennisdoomen, line 43: 🔧 This entire algorithm needs some documentation to explain the different parts.

dennisdoomen (changes requested):

src/Fallout.Migrate/Steps/CleanupBootstrapScriptsStep.cs
```diff
@@ -0,0 +1,62 @@
...
+                ? value
+                : Environment.NewLine;
+
+        var lines = GetLineFeedRegex().Split(content).ToList();
+
+        var indexOfEnterpriseEnvVarCheck = lines.FindIndex(line => line.Contains("NUKE_ENTERPRISE_TOKEN"));
+        var endOfIfBlock = lines.FindIndex(indexOfEnterpriseEnvVarCheck,
+            line => line.Trim() == "}" || line.Trim() == "fi");
+
+        if (lines[endOfIfBlock + 1].Trim() == "")
```
dennisdoomen: 🔧🤖 `endOfIfBlock` (line 47) can come back `-1` if the `if`/`fi`/`}` block is malformed, and `lines[endOfIfBlock + 1]` here throws `IndexOutOfRangeException` if the block is the last content in the file. This runs against arbitrary user-modified bootstrap scripts during migration, so a hand-edited or unusual file could crash the migration step instead of degrading gracefully (e.g. skip with a warning).

src/Fallout.Migrate/Steps/CleanupBootstrapScriptsStep.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        string newline =
+            GetLineFeedRegex().Match(content).Value is { Length: > 0 } value
+                ? value
+                : Environment.NewLine;
+
+        var lines = GetLineFeedRegex().Split(content).ToList();
+
+        var indexOfEnterpriseEnvVarCheck = lines.FindIndex(line => line.Contains("NUKE_ENTERPRISE_TOKEN"));
```
dennisdoomen: ♻️🤖 `nukeEnterprisePattern` (word-boundary regex) is only used for the initial `IsMatch` guard, while this line re-does the detection with a plain `.Contains("NUKE_ENTERPRISE_TOKEN")`. Could drop the regex field and just check `indexOfEnterpriseEnvVarCheck == -1` for the guard instead of two inconsistent checks.

src/Fallout.Migrate/Steps/CleanupBootstrapScriptsStep.cs
ITaluone: Do I got this correct, that the one integration test would suffice then?

ITaluone: I cannot see a way to not test the actual internals, without bloating this PR.

Should we for now proceed with this and refactor the tests later?

src/Fallout.Migrate/Steps/CleanupBootstrapScriptsStep.cs
ITaluone: Ahm.. what does this mean?

in principle: this for example is valid bash as well
```bash
 if [[ ! -z ${NUKE_ENTERPRISE_TOKEN+x} && "$NUKE_ENTERPRISE_TOKEN" != "" ]]; then
                                                             "$DOTNET_EXE" nuget remove source "nuke-enterprise" &>/dev/null || true
                                                             "$DOTNET_EXE" nuget add source "https://f.feedz.io/nuke/enterprise/nuget" --name "nuke-enterprise" --username "PAT" --password "$NUKE_ENTERPRISE_TOKEN" --store-password-in-clear-text &>/dev/null || true; fi
```

but just grepping for "fi" is potentially more dangerous..


the +1 thing is fixed btw..

src/Fallout.Migrate/Steps/CleanupBootstrapScriptsStep.cs
ITaluone: Ah.. nice finding

src/Fallout.Migrate/Steps/CleanupBootstrapScriptsStep.cs
ITaluone: I searched a bit..

I could give https://github.com/tree-sitter/tree-sitter/ (for bash) and the official MS powershell parser (`using System.Management.Automation.Language;`) a try.

@dennisdoomen Is this too much?

src/Fallout.Migrate/Steps/CleanupBootstrapScriptsStep.cs
dennisdoomen: What I meant is that for now, we should always use the public API of that step, and that is not that `internal` Cleanup.

src/Fallout.Migrate/Steps/CleanupBootstrapScriptsStep.cs
dennisdoomen: Yeah, I think that your fix is good enough for now. If somebody completely rewrote the script, they are out of luck

src/Fallout.Migrate/Steps/CleanupBootstrapScriptsStep.cs
ITaluone: Ok.. obviously I don't get what you asking for :)

How can I _not_ use the internal `Cleanup` in tests? Given, that every call to XxxStep needs a `MigrationContext`* and a `Summary` (passing `null` is not working).  
And to provide this, I basically would double the code which already exists in the `MigrationIntegrationSpecs`.

*) Which in return needs only the root directory, where the files supposed to be and not the content, makes it really "hard" to do that inline

So, I can think of three solutions for now
1) leave it as is for now, and restructure the tests to only use the `MigrationIntegrationSpecs`
2) remove all tests which do so right now (meaning: all of them), since `ScripRewriter.Rewrite` is exactly the same...
3) restructure now (and bloat this PR)

dennisdoomen: > Given, that every call to XxxStep needs a MigrationContext* and a Summary (passing null is not working). And to provide this, I basically would double the code which already exists in the MigrationIntegrationSpecs.

Creating a `MigrationContext` and `Summary` is pretty trivial, so I'm wondering why you think that makes it difficult. Yes, you would need a temporary `build.ps1`, but that's hardly an issue.

ITaluone: Ok, hope this fits now :)

dennisdoomen (approved):

dennisdoomen:
@ITaluone Just needs a rebase

ITaluone:
@dennisdoomen done

I refactor the tests tomorrow, to all use the public API of XxxxStep and the `AbsolutePath.Temp`

src/Fallout.Migrate/Steps/CleanupBootstrapScriptsStep.cs
ChrisonSimtian: yeah generally speaking, i'd rather have to go through the pain once to set things up properly but then down the line we're able to test under actual conditions than doing tests that bypass real conditions and dont give us the right conditions i.e. dont alert us when something breaks.

Gotta keep in mind, we're not a dev sweatshop, if we take our sweet time to do things right (and more importantly set a good examples for other devs AND for AI) then thats fine ;-) lets take that time and do things right :-)

ChrisonSimtian (approved):

ITaluone:
Is there any chance that this goes into 10.4?

dennisdoomen:
No, everything that landed on `main` after we created the `release` branch will end up on 10.5. But as soon as 10.4 is reliably out of the door, 10.5 can follow pretty quickly.

PR #528, opened 2026-07-21
Smooth the Fallout.Common.ProjectModel → Fallout.Solutions migration path

ChrisonSimtian:
Follow-up to #257 (the rename of `Fallout.Common.ProjectModel` to `Fallout.Solutions`). That rename shipped in 11.0.13 as a breaking change, with no compatibility shim. This PR does not undo that break — it stays. It adds a shim and a migration-tool fix so consumers upgrading from **11.0.1–11.0.12** have an easier, mechanical way to move to the new namespace.

Non-breaking / additive.

### What changed
- **Fix the CLI migrator rule.** `fallout-migrate` only replaced the `Nuke.` prefix, so `using Nuke.Common.ProjectModel;` was rewritten to the now-unused `Fallout.Common.ProjectModel`, not the real new namespace. It now maps both the old `Nuke.Common.ProjectModel` and the interim `Fallout.Common.ProjectModel` to `Fallout.Solutions` in one edit. This closes a gap in the CLI tool that #253 (an earlier fix) left open — #253 only fixed the analyzer's code-fix, not the CLI migrator.
- **Add a transition shim** (a small compatibility layer that keeps old code working against the new API) for the common entry-point pattern. `Solution` and `SolutionAttribute` are mirrored back into `Fallout.Common.ProjectModel` inside `Fallout.Common`, so `[Solution] readonly Solution Solution;` still compiles.
- **Add a compile-time check.** A new test project, `Fallout.Consumer.ProjectModelShim` (in `fallout.slnx`), fails CI if the shim above stops working. This mirrors how `Nuke.Consumer` already guards the `Nuke.Common.ProjectModel` shim.

### Limits
This shim only covers the common entry-point pattern above — it is not a full alias for the old namespace. Code that navigates the project graph still gets the real `Fallout.Solutions.*` types. The `GenerateProjects` code generator still keys off the new namespace's attribute. Binary compatibility and the package-ID rename are unaffected. For anything beyond the entry-point pattern, `fallout-migrate` is the complete fix — same as the existing `Nuke.Common.ProjectModel` shim.

### Verification
- `Fallout.Common`, the shim, and the new consumer sentinel build with 0 errors.
- `Fallout.Migrate.Specs` passes (34 tests, 4 new ones for the ProjectModel rule).


src/Fallout.Migrate/Steps/CodeRewriter.cs
```diff
@@ -31,7 +41,13 @@ public static RewriteResult Rewrite(string original)
...
         var edits = 0;
 
-        var content = namespacePrefix.Replace(original, _ =>
+        var content = projectModelNamespace.Replace(original, _ =>
+        {
+            edits++;
+            return "Fallout.Solutions";
+        });
+
+        content = namespacePrefix.Replace(content, _ =>
```
ITaluone: I understand the new migration design, that new migrations should go into their dedicated `XxxStep".

ChrisonSimtian: yep, claude doesn't quite understand it yet. so this needs some work plus some agent.md adjustments. just down with a pretty bad cold this week so doing things slower and one at a time 😊

ITaluone (approved):

dennisdoomen:
I was really struggling with the title and the description. Took quite some time until I understood that this is a fix for a previous breaking change on the `main` branch. The cryptic AI-generated description and all the references to v11 put me on the wrong foot. 

I've created https://github.com/Fallout-build/Fallout/pull/544 to help prevent that situation and took the liberty to have Copilot rewrite the description using those new instructions. Now I finally get it 😀

dennisdoomen (approved):

src/Fallout.Migrate/Steps/CodeRewriter.cs
ChrisonSimtian: reason why claude doesnt understand it is: we have a mix of old and new pattern, claude picked the old one...

ChrisonSimtian:
> I was really struggling with the title and the description. Took quite some time until I understood that this is a fix for a previous breaking change on the `main` branch. The cryptic AI-generated description and all the references to v11 put me on the wrong foot.
> 
> I've created #544 to help prevent that situation and took the liberty to have Copilot rewrite the description using those new instructions. Now I finally get it 😀

yeah not sure why but at some point in the last week or so the model has changed and now its pretending to be a uber-master-scholar or something ... gotta tame it back a bit :D

PR #533, opened 2026-07-23
Track the current majors for the actions the generator emits

lahma:
The GitHub Actions generator hard-codes the action versions it emits, so a consumer can't move any of them forward — the only escape is subclassing `GitHubActionsAttribute` and replacing the emitted steps. This moves all three to their current majors:

| Action | Before | After | Current major released |
|---|---|---|---|
| `actions/setup-dotnet` | v4 | **v6** | 2026-07-16 |
| `actions/checkout` | v6 | **v7** | 2026-06-18 |
| `actions/upload-artifact` | v5 | **v7** | 2026-02-26 |

`setup-dotnet@v4` additionally runs on the node20 runtime GitHub has scheduled for removal.

## Why it matters downstream

Two repos migrating off NUKE hit this:

- **Quartz.NET** ([quartznet/quartznet#3162](https://github.com/quartznet/quartznet/pull/3162)) carried a custom step class purely to reach `setup-dotnet@v5`, because NUKE emitted no SDK setup at all. Dropped on migration, accepting `@v4` as a temporary step back.
- **test262-harness-dotnet** carries custom step classes *today* to reach `checkout@v7` and `upload-artifact@v7`, with the stated intent of tracking action majors independently of the framework version. Those classes exist only because the generator's pins lag.

That's the argument for the generator tracking current majors: every consumer that wants them otherwise reimplements step emission, and those reimplementations then drift from the framework's own output.

## Scope

- the three emitting step classes
- 28 Verify snapshots asserting generated YAML
- `docs/05-cicd/github-actions.md` sample output
- this repo's own workflows, including the hand-written `publish-packages-*` ones (the generated ones would change on the next regeneration regardless)

One `checkout@v6` string remains in `GitHubActionsStepPipelineSpecs` — that's a user-supplied `GitHubActionsCustomStep` fixture, not a generator pin, so it's arbitrary sample input.

`dotnet test tests/Fallout.Common.Specs` — 131 passed, 0 failed.

### DIFF
--- a/.github/workflows/build-cross-platform.yml
+++ b/.github/workflows/build-cross-platform.yml
@@ -49,7 +49,7 @@ jobs:
             ~/.nuget/packages
           key: ${{ runner.os }}-${{ hashFiles('**/global.json', '**/*.csproj', '**/Directory.Packages.props') }}
       - name: 'Setup: .NET SDK'
-        uses: actions/setup-dotnet@v4
+        uses: actions/setup-dotnet@v5
         with:
           global-json-file: global.json
       - name: 'Restore: dotnet tools'
@@ -71,7 +71,7 @@ jobs:
             ~/.nuget/packages
           key: ${{ runner.os }}-${{ hashFiles('**/global.json', '**/*.csproj', '**/Directory.Packages.props') }}
       - name: 'Setup: .NET SDK'
-        uses: actions/setup-dotnet@v4
+        uses: actions/setup-dotnet@v5
         with:
           global-json-file: global.json
       - name: 'Restore: dotnet tools'
--- a/.github/workflows/build-cross-platform.yml
+++ b/.github/workflows/build-cross-platform.yml
@@ -38,7 +38,7 @@ jobs:
     name: macos-latest
     runs-on: macos-latest
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@v7
         with:
           fetch-depth: 0
       - name: 'Cache: .fallout/temp, ~/.nuget/packages'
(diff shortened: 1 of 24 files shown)


ITaluone (approved):

lahma:
Widened while this sat open, and the setup-dotnet target moved.

**setup-dotnet@v5 was already stale.** v6.0.0 shipped 2026-07-16, after this PR was opened against v5. Pushed the pin to v6 rather than landing a version that's superseded on arrival.

**Added checkout and upload-artifact**, the other two actions the generator pins — v6→v7 and v5→v7. Folded into this PR rather than a second one because both touch the same 28 Verify snapshots and would conflict.

Retitled accordingly; the branch name still says `setup-dotnet-v5`, which renaming would break the PR to fix, so it stays.

The motivating case is now concrete: test262-harness-dotnet carries custom step classes *today* solely to reach checkout@v7 and upload-artifact@v7. With this merged those classes delete outright.

Specs re-run: 131 passed, 0 failed.

ITaluone:
Should we setup some sort of "heads-up tests"? Meaning: something that notifies us, when a new release is out..

lahma:
> Should we setup some sort of "heads-up tests"? Meaning: something that notifies us, when a new release is out..

I think easiest option would be to allow user's to override the action name/version via attribute property, then lagging a bit behind wouldn't stop users from getting the latest by updating their build.

ChrisonSimtian:
> > Should we setup some sort of "heads-up tests"? Meaning: something that notifies us, when a new release is out..
> 
> I think easiest option would be to allow user's to override the action name/version via attribute property, then lagging a bit behind wouldn't stop users from getting the latest by updating their build.

I agree. Should we get this one out to bump versions and at the same time open a gh issue to track this as a long-term solution?
Also, welcome to Fallout :-)

ChrisonSimtian (approved):

ITaluone:
> Should we get this one out to bump versions and at the same time open a gh issue to track this as a long-term solution?

Either one is fine, but "separation of concerns"-wise, yes :)

PR #531, opened 2026-07-23
Configure Dependabot for weekly dependency updates

YousifSala7:
Closes #6

**Description:**
Added Dependabot configuration to automate dependency updates for NuGet, .NET SDK, and GitHub Actions ecosystems.

**Changes:**
* Configured the `nuget` ecosystem to monitor `Directory.Packages.props`.
* Configured the `dotnet-sdk` ecosystem to monitor `global.json`.
* Configured the `github-actions` ecosystem to monitor `.github/workflows/`.
* Set `schedule` to `weekly` and utilized `groups` for `nuget` and `github-actions` to bundle version updates into a single PR per ecosystem.
* Documented the dependency update strategy in `AGENTS.md` (as `CLAUDE.md` delegates its context there).

PR #537, opened 2026-07-24
Configure Dependabot for GitHub Actions updates

YousifSala7:
Description:

Follow-up to #531.
Adds Dependabot configuration for the `github-actions` ecosystem.
Cross-references #535, which introduces configurable action references for generated GitHub Actions workflows.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,3 +15,13 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+
+  # Monitor GitHub Actions (.github/workflows/*.yml)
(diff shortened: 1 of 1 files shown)


ITaluone (approved):
Cool

Despite using dependabot for things generated elsewhere...

dennisdoomen (approved):

ChrisonSimtian (approved):
yes but like i said in the last pr, this will get downgraded again unless we change the build as well. i doubt dependabot understands that pattern, so HITL will have to touch those PRs and upgrade both.

But fine by me lets do it and see if its more noise than use or not 😅

PR #541, opened 2026-07-24
Reintroduce JetBrains [PublicAPI] as an IDE authoring hint

### LINKED ISSUE #539: Reintroduce JetBrains [PublicAPI] as an IDE authoring hint on consumer-facing assemblies
## Problem

[PR #76](https://github.com/Fallout-build/Fallout/pull/76) removed `JetBrains.Annotations` wholesale (~1450 `[PublicAPI]` + friends). Side effect: Rider/ReSharper now flag Fallout's public **build-authoring surface** as "unused" — because it *is* unused in-repo by design (external consumers call it, we don't). Contributors get persistent false-positive "unused symbol" noise on exactly the API we ship.

`[PublicAPI]` tells the IDE "this is intentional API, consumed externally — don't flag it." It's an **IDE authoring hint only**: zero runtime/semantic effect.

**Not this issue:** API-surface break detection / changelog gating. That's #410 / #530 (Roslyn `PublicApiAnalyzers`). Complementary and separate — one silences an IDE hint, the other guards the shipped surface.

## Outcome

Consumer-facing ("outer") assemblies advertise their public surface as intentional API again, so:

- contributors stop getting false unused-symbol warnings on shipped API;
- consumers' own Rider/ReSharper sees our surface as API too (the annotation flows through the package, no `PrivateAssets`).

## Acceptance criteria

- [ ] `JetBrains.Annotations` referenced so the annotation flows to consumers (no `PrivateAssets`).
- [ ] Outer/consumer-facing assemblies carry `[assembly: PublicAPI]` via a **single allowlist** (one source of truth), not ~1450 hand-placed attributes.
- [ ] Inner assemblies (Cli, Migrate, SourceGenerators, MSBuildTasks, Tooling.Generator, Persistence) get **no blanket**; genuine extensibility types there are annotated per-type (`[PublicAPI]` in source) — the hybrid split.
- [ ] An ADR records the decision (mechanism + outer/inner criterion) and goes up for public review.
- [ ] An encapsulation-audit doc tracks any public-by-accident types the pass surfaces.

ChrisonSimtian:
Closes #539. Reintroduces `[PublicAPI]` for the IDE "unused symbol" hint only — **not** API-break detection (that's the #530 / #540 track, a separate concern).

Rider/ReSharper flags Fallout's public build-authoring surface as unused because it *is* unused in-repo by design (external consumers call it, we don't). `[PublicAPI]` marks it as intentional API. IDE hint only — no runtime/semantic effect.

## Approach — hybrid (see [ADR-0011](docs/adr/0011-reintroduce-publicapi-ide-hint.md))

- **Outer/consumer-facing assemblies** get a blanket `[assembly: PublicAPI]` via the `_FalloutOuterApiAssemblies` allowlist in `Directory.Build.props` (single source of truth). The `JetBrains.Annotations` reference flows to consumers (no `PrivateAssets`), so their Rider sees our surface as API too.
- **Inner assemblies** get no blanket. The four `Fallout.Persistence.Solution` model types that surface through `Fallout.Solution`'s public API (`SolutionModel`, `SolutionItemModel`, `SolutionProjectModel`, `SolutionFolderModel`) carry a per-type `[PublicAPI]` instead — a blanket would bless the vendored parser's implementation detail.
- Opt out per project with `<FalloutPublicApi>false</FalloutPublicApi>`.

Only `[PublicAPI]` comes back; the other JetBrains attributes removed in #76 stay removed.

## Docs

- `docs/adr/0011-reintroduce-publicapi-ide-hint.md` — **Proposed**, up for discussion here.
- `docs/api-encapsulation-audit.md` — the outer/inner split, the per-type carve-outs, and the public-by-accident follow-up the coarse blanket implies.

Full solution builds green; `[assembly: PublicAPI]` verified emitted on outer assemblies and absent on inner.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -69,4 +69,26 @@
     <Compile Include="$(MSBuildThisFileDirectory)AssemblyInfo.cs" />
   </ItemGroup>
 
+  <!-- JetBrains.Annotations [PublicAPI] — an IDE authoring hint only (no runtime/semantic effect); see
+       docs/adr/0011-reintroduce-publicapi-ide-hint.md. The "outer" (consumer-facing) layer gets a blanket
+       [assembly: PublicAPI] so Rider/ReSharper treat its entire public surface as intentional API and never
+       flag it as unused (it is unused *in-repo* by design — external consumers call it, we don't).
+       _FalloutOuterApiAssemblies is the single source of truth for that layer. Inner layers (Cli, Migrate,
+       SourceGenerators, MSBuildTasks, Tooling.Generator, Persistence.Solution) are NOT blanket-marked; any
+       genuine extensibility type there is annotated with [PublicAPI] in source (the hybrid split — see the
+       ADR and docs/api-encapsulation-audit.md). A project opts out with <FalloutPublicApi>false</FalloutPublicApi>
+       set before this import. This is NOT public-API break detection — that is the Roslyn PublicApiAnalyzers
+       track (#410 / #530), a separate concern. -->
+  <PropertyGroup>
+    <_FalloutOuterApiAssemblies>;Fallout.Common;Fallout.Build;Fallout.Build.Shared;Fallout.Components;Fallout.Core;Fallout.Tooling;Fallout.ProjectModel;Fallout.Utilities;Fallout.Utilities.IO.Compression;Fallout.Utilities.IO.Globbing;Fallout.Utilities.Net;Fallout.Utilities.Text.Json;Fallout.Utilities.Text.Yaml;Fallout.Solution;Nuke.Common;Nuke.Build;Nuke.Components;</_FalloutOuterApiAssemblies>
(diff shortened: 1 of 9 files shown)


ChrisonSimtian:
@dennisdoomen i brought `[PublicAPI]` back but this time not a shotgun make everything publicapi, instead its a bit more targeted both via Attributes and via Assemblies (mind you still too many for my taste but once we make it through a few refactorings hopefully the surface will normalise a bit)

ITaluone (approved):
Nice!

Never got the reason for removing in the first place 🫢  But cool to see the actual value ;)

ChrisonSimtian:
> Never got the reason for removing in the first place 🫢

1. I was missing a narrative for why we use them (which this PR will fix with an ADR and a consistent gh issue story)
2. I felt they were just wildly applied everywhere in the project without sense

So this time its a bit more targeted, organised and I want the AI instructions side-by-side to ensure we maintain those properly.

PR #542, opened 2026-07-26
Add weekly prune of old -preview package versions

ChrisonSimtian:
Adds a weekly workflow that prunes old `-preview` package versions from GitHub Packages, keeping the newest 20 per package.

## Problem

GitHub Packages has no retention-policy feature. There is no TTL and no auto-expiry. `publish-packages-preview.yml` pushes the full package set on every `main` commit ([ADR-0004](../blob/main/docs/adr/0004-calendar-versioning-and-dual-pace-channels.md)), so versions accumulate without limit. The feed now holds **2,002 preview versions across 24 packages**.

## Change

Weekly, and on demand, prune keeping the newest **20** `-preview` versions per package.

- **Protects** GA, `-rc.N` and legacy `-alpha.N` through `ignore-versions`. Only per-commit previews are eligible.
- **Discovers** packages from the API instead of a hard-coded list, so a new package is covered the moment it first publishes.
- **Dry-run by default** on manual dispatch, with a per-package report. Scheduled runs prune for real.
- Orders by creation date, not version number. The feed holds two preview schemes (`2026.1.0-preview.<height>` and `10.0.0-preview.<height>`), so a version-number sort would keep the wrong ones.

Measured against the live feed: prunes **1,553** versions and leaves 449 previews, plus the 5 protected versions on each package. The largest single-package prune is 71 versions, which is under the action's limit of 100 deletions per run.

## Notes

- **The original storage reason no longer applies.** This PR was written when 22 of 23 packages were private and over the 500 MB free-plan quota. All 24 packages are public now, so their storage is free and the organisation's net spend is $0. This change is now hygiene to stop unbounded growth, not a quota fix. The comment header was updated to say so.
- Deletion is irreversible. Dispatch once with `dry-run=true` before letting the schedule run.
- The report step re-implements the keep-newest-N count in shell rather than calling the action. Treat it as a close estimate, not a byte-exact promise.
- **Unrelated bug spotted:** `version.json` on `main` is `10.0.0-preview.{height}`, so `main` publishes `10.0.0-preview.18`. The preview version went backwards from `2026.1.0-preview.156` and no longer matches the calendar-version scheme in AGENTS.md and ADR-0004. Worth a separate issue.

### DIFF
--- a/.github/workflows/prune-preview-packages.yml
+++ b/.github/workflows/prune-preview-packages.yml
@@ -1,0 +1,141 @@
+# Hand-written. Storage hygiene for the `-preview` channel.
+#
+# Why this exists: GitHub Packages has NO retention policy feature — no TTL, no
+# auto-expiry, nothing to configure. Package versions live forever until something
+# deletes them. publish-packages-preview.yml pushes the full package set to GitHub
+# Packages on EVERY `main` commit (ADR-0004), so versions accumulate without bound;
+# by August 2026 that had reached ~2,280 versions across 24 packages. Every package
+# is public, and public package storage is free, so there is no quota pressure today.
+# This prune is hygiene: it stops the feed growing without limit. Actions artifacts
+# are not the problem (they DO have retention — see `retention-days` in
+# publish-packages-release.yml).
+#
+# Shape:
+#   - Discovers the org's NuGet packages from the API rather than hard-coding a
+#     list, so a newly-added package is covered the moment it first publishes.
+#   - Prunes ONLY `-preview.` versions, keeping the newest `keep` of them.
+#   - `ignore-versions` protects everything that is not a per-commit preview:
+#     GA, `-rc.N`, and the legacy `-alpha.N` versions are never touched. Those are
+#     the durable release markers and must outlive any cleanup.
+#   - Runs weekly, matching the Dependabot cadence, and on demand.
+#
+# "Newest" means most-recently-created, not highest version number: the action
+# orders by CREATED_AT descending. That matters because the feed currently holds
+# two different preview schemes (`2026.1.0-preview.<height>` and the lower-sorting
+# `10.0.0-preview.<height>` that `main` publishes today), so a version-number sort
+# would keep the wrong ones. Creation order is correct regardless of scheme.
+#
+# Deletion is IRREVERSIBLE. Run it via workflow_dispatch with dry-run=true first;
+# the report step prints exactly what a real run would remove.
+#
+# Note: preview versions are disposable by design — they are per-commit builds of
+# the integration trunk, reproducible from the commit in the version string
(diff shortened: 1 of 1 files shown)


ChrisonSimtian:
Update: the 22 private packages have now been flipped to **public** (all 23 NuGet packages verify as \`public\` via the API). Billed Packages storage is ~0, so the org is out of the overage.

That changes this PR from *quota-critical* to *hygiene*: it is no longer needed to get under the limit, but still worth having — ~1,650 per-commit preview versions is unmanageable noise, and any **new** package will publish as private by default, which is exactly how this went unnoticed.

Two corrections to the description above:
- Package settings live at `/orgs/{org}/packages/nuget/{name}/settings` — **no** `/package/` segment (that shape 404s). The package page itself is `/{org}/{repo}/pkgs/nuget/{name}`.
- "not scriptable here" was wrong about the UI being the only option in practice — it is UI-only in the sense that no REST/GraphQL mutation exists (GraphQL exposes only `deletePackageVersion`), but the flow automates fine via Playwright: Change visibility → Public → type package name to confirm → submit.

ChrisonSimtian:
leaving this open for now, I made those packages public so problem solved for now but at some point I'd like to tidy up and have some housekeeping.
Would also be cool to have this as a fallout based workflow, not a claude hand cranked the yml file workflow

ChrisonSimtian:
Reviewed and rebased onto `main` (now includes the SDK pin from #640). Verified against the live feed before merging.

**Checked and correct**

- `ignore-versions: ^(?!.*-preview\.).*$` protects exactly the right set. Tested against real version strings from the feed: `10.4.0`, `2026.1.0-rc.1` and `10.0.0-alpha.4` are all ignored; `10.5.0-preview.2.gace05f3695` and `2026.1.0-preview.156.g…` are eligible.
- `min-versions-to-keep` takes precedence over `num-old-versions-to-delete`, so the default of 1 does not apply here.
- The action deletes at most 100 versions per run. The largest single-package prune is 71, so one run is enough per package.
- Measured impact: 2,002 preview versions, 1,553 pruned, 449 left, and the 5 protected versions on each package untouched.
- Action pinned to a major tag (`@v5`), matching the convention used by the other workflows, and Dependabot already covers `github-actions`.

**Changed while rebasing**

- Header comment said the prune fixed a quota problem. All 24 packages are public now, so that storage is free and net spend is $0. Reworded to say this is hygiene against unbounded growth.
- The report step described itself as "the same arithmetic the prune job applies". It re-implements the count in shell rather than calling the action, so the comment now says it is a close estimate.
- Stale count "the other 22" corrected.

**Follow-ups, not blocking**

- The `${{ steps.list.outputs.packages }}` interpolation inside `run:` would be better passed through `env:`. Package names come from our own API so the risk is low, and there is no workflow linter in CI today.
- Dispatch once with `dry-run=true` before the first Monday schedule fires. Deletion is irreversible.

PR #544, opened 2026-07-26
Make the non-native-English PR/issue guidance concrete

dennisdoomen:
PR #528's description is a good example of what the existing "write for non-native English readers" rule was meant to prevent, but the rule was one abstract line and easy to skip.

### What changed
- Breaks the rule in `issue-and-pr-style.md` into checkable guidance: one idea per sentence, no idioms/figurative language, define jargon on first use, gloss cross-references, prefer short words.
- Adds two before/after anti-pattern rows using real sentences from #528.
- Adds a glossary of repo jargon (shim, sentinel, canonical type, ceiling, etc.) to `conventions.md`, linked from the rule above.

### DIFF
--- a/docs/agents/conventions.md
+++ b/docs/agents/conventions.md
@@ -27,6 +27,23 @@ Applies to AI tools and humans. Many readers are non-native English speakers —
 
 Covers GitHub issues, PR titles/bodies/comments, and commit messages.
 
+## Glossary of repo jargon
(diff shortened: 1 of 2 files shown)


ChrisonSimtian:
lets take it for a test ride aye?

ChrisonSimtian (approved):

PR #559, opened 2026-07-26
Emit real blank lines instead of a zero-width space

### LINKED ISSUE #551: Console output emits invisible U+200B and U+00A0 characters
### Problem
Two hacks put non-printing characters into normal build output:

- `AnsiConsoleHostTheme.Format` returns `"​"` (zero-width space) instead of `""` for blank text ([src/Fallout.Build/Theming/AnsiConsoleHostTheme.cs:110](https://github.com/Fallout-build/Fallout/blob/main/src/Fallout.Build/Theming/AnsiConsoleHostTheme.cs#L110)). Every blank line in Fallout output is a ZWSP line.
- `Host.WriteLogo` does `x.Replace(" ", " ")` ([src/Fallout.Build/Host.cs:54](https://github.com/Fallout-build/Fallout/blob/main/src/Fallout.Build/Host.cs#L54)), turning every space in the logo into a non-breaking space.

`SystemConsoleHostTheme.Format` returns the text unchanged for the same input, so the two themes disagree — the ZWSP is not a deliberate cross-theme behaviour.

Consequences: `grep "^$"` does not match blank lines in build logs, copied output carries invisible characters into issues and chat, and log diffing/snapshotting sees noise.

### Outcome
Blank output lines are actually blank. The logo uses ordinary spaces.

### Acceptance criteria
- [ ] `IHostTheme.Format*` returns an empty string for null/empty/whitespace input in both themes
- [ ] Rendered logo lines contain no U+00A0
- [ ] Specs assert no U+200B or U+00A0 in the emitted output

ChrisonSimtian:
Removes the non-printing characters from normal build output.

### What changed
- Both host themes return an empty string for null, empty, or whitespace-only text. They previously returned `U+200B` (zero-width space), so every blank line of build output carried an invisible character.
- `SystemConsoleHostTheme.Format*` routes through the same normalisation. It used to return the input unchanged, so `null` in gave `null` out, and the two themes disagreed.
- `Host.WriteLogo` no longer rewrites every space as `U+00A0` (non-breaking space).
- The blank-text contract is now written on `IHostTheme`.

### Why
`grep "^$"` did not match blank lines in build logs. Copied output carried invisible characters into issues and chat.

The issue named one zero-width-space site. There were two: `SystemConsoleHostTheme.Write` had the same substitution.

Whitespace-only text now renders as an empty line in the system theme, where it previously kept its spaces. That matches what the ANSI theme already did, so the two themes agree.

One risk worth a reviewer eye: the non-breaking spaces may have been intended to stop an HTML-rendering CI log viewer from collapsing the logo indentation. Modern viewers render pre-formatted text, so ordinary spaces should be fine. A spec pins the tagline indentation so a regression here is caught.

### Verification
`dotnet test tests/Fallout.Build.Specs` and `tests/Fallout.Common.Specs`. 13 new specs pass, 12 of which failed before the fix.

Closes #551

### DIFF
--- a/tests/Fallout.Build.Specs/ConsoleOutputCharacterSpecs.cs
+++ b/tests/Fallout.Build.Specs/ConsoleOutputCharacterSpecs.cs
@@ -1,0 +1,102 @@
+using System;
+using System.IO;
+using System.Linq;
+using Fallout.Common.Execution.Theming;
+using FluentAssertions;
+using Xunit;
+
+namespace Fallout.Common.Specs;
+
+/// <summary>
+/// Guards normal console output against non-printing characters. See #551. Two hacks put them
+/// there: the ANSI theme returned a zero-width space instead of an empty string for blank text, and
+/// <see cref="Host.WriteLogo"/> rewrote every space in the logo as a non-breaking space.
+/// </summary>
+[Collection(ProcessGlobalStateCollection.Name)]
+public class ConsoleOutputCharacterSpecs
+{
+    private const string ZeroWidthSpace = "​";
+    private const string NonBreakingSpace = " ";
+
+    /// <summary>Leading spaces on the logo's tagline lines.</summary>
+    private const int TaglineIndent = 21;
+
+    public static TheoryData<string> BlankTexts => new() { null, string.Empty, " ", "   " };
+
+    [Theory]
+    [MemberData(nameof(BlankTexts))]
+    public void The_ansi_theme_renders_blank_text_as_an_empty_string(string text)
+    {
+        var theme = (IHostTheme)AnsiConsoleHostTheme.Default256AnsiColorTheme;
+
+        theme.FormatInformation(text).Should().BeEmpty();
+    }
+
+    [Theory]
+    [MemberData(nameof(BlankTexts))]
+    public void The_system_theme_renders_blank_text_as_an_empty_string(string text)
+    {
+        var theme = (IHostTheme)SystemConsoleHostTheme.DefaultSystemColorTheme;
+
+        theme.FormatInformation(text).Should().BeEmpty();
+    }
+
+    [Fact]
+    public void The_two_themes_agree_on_blank_text()
+    {
+        var ansi = (IHostTheme)AnsiConsoleHostTheme.Default256AnsiColorTheme;
+        var system = (IHostTheme)SystemConsoleHostTheme.DefaultSystemColorTheme;
+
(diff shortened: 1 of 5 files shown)


ChrisonSimtian:
this one is interesting, as it mostly affects AI agents only, who would ever grep a buildoutput? Its fire and forget (but not anymore with AI, is it?)

tests/Fallout.Build.Specs/ConsoleOutputCharacterSpecs.cs
```diff
@@ -0,0 +1,114 @@
...
+    public void The_logo_still_indents_its_tagline()
+    {
+        var tagline = StripAnsi(CaptureLogo())
+            .Split('\n')
+            .Single(x => x.Contains("survived the NUKE"));
+
+        tagline.Should().StartWith("                     ");
+    }
+
+    private static string StripAnsi(string text) => Regex.Replace(text, @"\x1b\[[0-9;]*m", string.Empty);
```
dennisdoomen: ⛏️I would not use a private method like that.

tests/Fallout.Build.Specs/ConsoleOutputCharacterSpecs.cs
```diff
@@ -0,0 +1,114 @@
...
+
+    private static string CaptureLogo()
+    {
+        var original = Console.Out;
+        // WriteLogo writes through the static Host.Debug, which resolves the theme from
+        // Host.Instance — and constructing any Host reassigns it. Restore it so this spec doesn't
+        // leak a live console-writing host into the rest of the collection.
+        var instanceProperty = typeof(Host).GetProperty("Instance", ReflectionUtility.Static)
```
dennisdoomen: 🤔 I'm not a fan of doing this kind of stuff. Isn't there a cleaner way?

dennisdoomen (approved):

tests/Fallout.Build.Specs/ConsoleOutputCharacterSpecs.cs
ChrisonSimtian: thats a claude special :D absolutely agree, AI got a bit creative here

dennisdoomen:
> this one is interesting, as it mostly affects AI agents only, who would ever grep a buildoutput? Its fire and forget (but not anymore with AI, is it?)

I regularly do, especially if I see weird things happening or get complicated run-time errors. Or is that not what you meant?

dennisdoomen (approved):

PR #560, opened 2026-07-26
Draw the target block banner as a box

### LINKED ISSUE #552: Target block banner rules are asymmetric and mis-measured
### Problem
`Host.WriteBlock` draws the top rule at `text.Length + 5` and the bottom rule at `max(text.Length - 4, 2)` ([src/Fallout.Build/Host.cs:72-74](https://github.com/Fallout-build/Fallout/blob/main/src/Fallout.Build/Host.cs#L72-L74)). The two widths are unrelated, so the banner is not a box:

```
╬═══════════
║ Levels
╬══
```

`text.Length` is also wrong for multi-line block text — the method splits on newlines to print the content, but measures the whole string including the newline characters, so a multi-line banner gets a rule far wider than its widest line.

### Outcome
Both rules span the same width, sized to the widest content line.

### Acceptance criteria
- [ ] Top and bottom rules are the same width for any block text
- [ ] Rule width is derived from the widest content line, not the raw string length
- [ ] Multi-line block text is covered by a spec

ChrisonSimtian:
Makes the target block banner symmetric. Before, the two rules had unrelated widths.

### What changed
- Size both rules from the widest content line plus the width of the `║ ` prefix.
- Measure the raw text lines, not the themed ones. Theming adds escape sequences that occupy no console cells.
- Add `BlockBannerGeometrySpecs` over single-line and multi-line block text.

### Why
The top rule was `text.Length + 5` and the bottom was `max(text.Length - 4, 2)`. Multi-line text was measured by the raw string length, which includes the newline characters, so a tall banner got a rule far wider than its widest line.

Before and after, for a target named `Levels`:

```
╬═══════════        ╬═══════
║ Levels            ║ Levels
╬══                 ╬═══════
```

### Verification
`dotnet test tests/Fallout.Build.Specs` — 10 new specs pass, 9 of which failed before the fix. Also confirmed against a real build run.

Closes #552

### DIFF
--- a/src/Fallout.Build/Host.cs
+++ b/src/Fallout.Build/Host.cs
@@ -61,17 +61,20 @@ public partial class Host
     {
         return DelegateDisposable.CreateBracket(() =>
         {
-            var formattedBlockText = text
+            var blockTextLines = text
                 .Split(new[]
                 {
                     EnvironmentInfo.NewLine
-                }, StringSplitOptions.None)
-                .Select(Theme.FormatInformation);
+                }, StringSplitOptions.None);
+
+            // The rule spans the widest content line plus the "║ " prefix. Measure the raw lines,
+            // not the themed ones — theming adds escape sequences that occupy no console cells.
+            var rule = "╬" + '═'.Repeat(blockTextLines.Max(x => x.Length) + 1);
 
             Debug();
-            Debug("╬" + '═'.Repeat(text.Length + 5));
-            formattedBlockText.ForEach(x => Debug($"║ {x}"));
-            Debug("╬" + '═'.Repeat(Math.Max(text.Length - 4, 2)));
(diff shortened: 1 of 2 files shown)


dennisdoomen:
Is this a bug or a stylistic improvement? AFAIK, the three character ╬══  was by design.

ChrisonSimtian:
> Is this a bug or a stylistic improvement? AFAIK, the three character ╬══ was by design.

no you're right, not a bug but a feature/enhancement.

I'm not quite sure where to go with the box design in general but for now I wanna strip some of the "complexity" out of it.
Long term I'd like to hand the style part over to spectre, they do the UI side really well and got some pretty good looking (and more importantly properly rendering) UI elements

dennisdoomen (approved):

PR #563, opened 2026-07-26
Promote main onto the 10.4 release line and fix release version stamping

ChrisonSimtian:
Promotes the work merged to `main` since the 10.4 line was cut onto `release/v10.4`, so the next rc carries it. Cut target: **`v10.4.0-rc.4`**.

19 commits promoted verbatim (cherry-picked, linear — `bf3b517f` skipped as an already-applied duplicate of `eabd043c`). Each was reviewed on its own PR against `main`; they carry their own changelog labels.

Plus two release-mechanics fixes found while preparing the cut.

### Release packages were stamped with a git-height suffix

`v10.4.0-rc.3` shipped 22 packages named `10.4.0-rc.3.geabd043cc2` instead of `10.4.0-rc.3`. The pipeline checks out the tag, so HEAD is detached and matches none of `version.json`'s `publicReleaseRefSpec` entries — they're all branch refs, and NB.GV never tests that spec against `refs/tags/*`. Setting `PublicRelease` on the Test + Pack step is the override that works; adding a `refs/tags` pattern to the refspec does nothing.

This matters most at GA, where a stable tag would otherwise publish with a prerelease-shaped suffix and sort below the version it claims to be.

### The rc number was a side effect of commit count

`10.4.0-rc.{height}` tied the rc number to how many commits a promotion carried — this 19-commit promotion took the next rc from `rc.3` to **`rc.23`**. Correctly ordered, but meaningless.

Now pinned literally (`10.4.0-rc.4`), so the number moves when we decide to cut rather than when we merge. The git height still feeds the fourth assembly-version component. `main` keeps `{height}` — per-commit previews want a value that always moves on its own.

Trade-off, called out in the runbook: tagging twice without bumping republishes an existing version, and `--skip-duplicate` swallows that silently. Check the number before tagging.

### Reviewer notes

- **Needs `Rebase and merge`** (the repo's only merge button) — the promotion is deliberately linear.
- `Test` fails 3 `GitRepositoryWorktreeSpecs` cases *locally on a fork*: the spec compares two `GitRepository` resolutions of one clone, which diverge only when `origin` isn't the branch's tracking remote. Confirmed green on CI here — 156/156 in `Fallout.Build.Specs` versus 153/156 locally.
- This rc is the **first exercise of `upload-artifact@v7` + `download-artifact@v8`** in the release pipeline (from the Dependabot group bump). No other workflow uses artifacts, so the pairing is untested here. Both are current latest majors; publish jobs are idempotent (`--skip-duplicate`) if a re-run is needed.
- Verified locally: `Pack` produces a clean `10.4.0-rc.4` across all 22 packages.
- nuget.org is untouched — this tag publishes to GitHub Packages + a GitHub Release only.


ChrisonSimtian:
RC version is `{Height}` so this just got a MASSIVE bump. That was honestly not my intention, that needs fixing at some point. But its a RC so its nit

Issue #557, opened 2026-07-26
Define the console glyph vocabulary: replace kaomoji with a width-aware icon set


ChrisonSimtian:
### Problem
The inherited NUKE glyph vocabulary is arbitrary. The build outcome line ends in kaomoji ([src/Fallout.Build/Host.cs:176-179](https://github.com/Fallout-build/Fallout/blob/main/src/Fallout.Build/Host.cs#L176-L179)):

```
Build succeeded on 26/07/2026 22:49:03. ＼（＾ᴗ＾）／
Build failed on 26/07/2026 22:49:45. (╯°□°）╯︵ ┻━┻
```

Elsewhere: `☢` in the logo tagline (deliberate, keep), `»` as a selection marker, `¬` as a confirmation marker (#555). Target status in the summary table is carried by **colour alone** — green `Succeeded`, yellow `NotRun`, red `Failed` — with no glyph.

There is no vocabulary: no rule for which glyph means what, no fallback for terminals that cannot render one, and no width handling.

### Outcome
A defined, documented icon set where each glyph is self-explaining, renders at a known width, and degrades to ASCII when the terminal cannot show it. Status is legible without colour.

### Acceptance criteria
- [ ] A single internal source of truth maps semantic states (succeeded / failed / skipped / not-run / aborted / warning / running) to glyphs
- [ ] Every glyph has an ASCII fallback, selected by a capability probe rather than assumed
- [ ] Glyph rendering goes through display-width measurement, not `string.Length`
- [ ] Kaomoji removed from the build outcome line
- [ ] `NO_COLOR` is honoured, and status stays distinguishable with colour off
- [ ] The vocabulary is documented so tool wrappers and plugins use the same glyphs

### Design constraints

**1. Column alignment breaks twice over.** `Host.WriteTargetOutcome` pads with `PadRight`/`PadLeft` over `string.Length` — UTF-16 code units. Emoji break both halves of that assumption:

| Glyph | Code units | Display cells |
|---|---|---|
| `✅` U+2705 | 1 | 2 |
| `⚠️` U+26A0 U+FE0F | 2 | 2 |
| `✔` U+2714 | 1 | 1 |
| `☢` U+2622 | 1 | 1 |

`PadRight` over-pads narrow glyphs and under-pads wide ones. This is why the item belongs to the Spectre migration: Spectre.Console measures display width and owns the table layout. Doing it on the current `PadRight` code means hand-rolling a width table.

[...]

PR #564, opened 2026-07-26
Mark an accepted prompt answer with a check instead of ¬

### LINKED ISSUE #555: Interactive prompts confirm a selection with ¬ (negation sign)
### Problem
`ConsoleUtility` marks a confirmed prompt answer with the negation sign ([src/Fallout.Build/Utilities/ConsoleUtility.cs:19](https://github.com/Fallout-build/Fallout/blob/main/src/Fallout.Build/Utilities/ConsoleUtility.cs#L19)):

```csharp
private static string Confirmed => "¬";
```

It is printed in green after `PromptForInput` and `PromptForChoice` accept a value — so the glyph that means "not" is used to mean "accepted". Users hitting `fallout :setup` see it on every answer.

### Outcome
A confirmed answer is marked with a glyph that reads as confirmation.

### Acceptance criteria
- [ ] Confirmed prompt lines are prefixed with a check mark, not `¬`
- [ ] The glyph is single-cell so the prompt lines stay aligned
- [ ] Specs cover the confirmation prefix for both `PromptForInput` and `PromptForChoice`

### Notes
Keep it to a single-width glyph — the prompt writers pad to `BufferWidth` on raw character count, so a double-width glyph would shift the line. The broader kaomoji/icon vocabulary is separate (sub-issue of #391).

ChrisonSimtian:
The interactive prompts confirmed an accepted answer with `¬`, the negation sign. The glyph that means "not" was used to mean "accepted".

### What changed
- `ConsoleUtility.Confirmed` is now `U+2713 CHECK MARK`.
- Add `PromptConfirmationGlyphSpecs`, which drives `PromptForInput` and `PromptForChoice` against a fake `IConsole` that answers by pressing Enter.

### Why
Users see the marker on every answer during `fallout :setup`.

The glyph has to occupy a single column. Both prompt writers pad to `BufferWidth` on raw character count, so a double-width glyph would push the line past the buffer. `U+2713` is single-cell and not a surrogate pair, and a spec pins both properties. Choosing the wider icon vocabulary, the capability probe, and the ASCII fallback is #557.

### Verification
`dotnet test tests/Fallout.Build.Specs` — 5 new specs pass, 4 of which failed before the fix.

Closes #555

### DIFF
--- a/src/Fallout.Build/Utilities/ConsoleUtility.cs
+++ b/src/Fallout.Build/Utilities/ConsoleUtility.cs
@@ -14,7 +14,9 @@ public class ConsoleUtility
 
     private static string Default => "[default: {0}]";
 
-    private static string Confirmed => "¬";
+    // Single-cell check mark: the prompt writers pad to BufferWidth on raw character count, so a
+    // wider glyph would push the line past the buffer. Glyph vocabulary at large is #557.
+    private static string Confirmed => "✓";
 
     private static string Selected => "»";
 
--- a/tests/Fallout.Build.Specs/PromptConfirmationGlyphSpecs.cs
+++ b/tests/Fallout.Build.Specs/PromptConfirmationGlyphSpecs.cs
@@ -1,0 +1,109 @@
+using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
+using Fallout.Build.Utilities;
+using Fallout.Common.Utilities;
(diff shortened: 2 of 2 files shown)


ChrisonSimtian:
tldr; use a checkmark for ticking off boxes not a ... whatever that is sign

dennisdoomen (approved):

Issue #104, opened 2026-05-22
AI-friendly default console output (terse/structured mode)


ChrisonSimtian:
## Problem

Default build output is verbose and human-oriented — banners, ANSI colour, per-tool stdout firehose, repeated headers. When an AI agent (Claude Code, Copilot CLI, Cursor, etc.) runs `./build.ps1`, most of the output is noise that burns context tokens without adding signal. Humans tolerate it; agents pay for every line.

## Proposal

Make the default output mode terser and more structured. Add an explicit mode switch (flag and/or env var) to opt back into the rich rendering for interactive use.

Strawman:

- `FALLOUT_OUTPUT=human|ai|json` env var, plus `--output human|ai|json` CLI flag.
- `ai` (default in non-TTY / `CI=true`): one line per target with status + duration, suppressed tool stdout, surfaced only on failure. No ANSI. Stable, greppable shape.
- `human` (default in interactive TTY): today's rendering — banners, colour, streamed tool output.
- `json` (opt-in): line-delimited JSON for tools.

## Why now

v12 is already a refactor wave (plugin SDK foundation + extension-point catalogue from [RFC #2](https://github.com/ChrisonSimtian/Fallout/issues/98)). The output-sink extension point is one of the listed catalogue items; this issue is the consumer-facing surface of that work.

## Open questions

- Default-mode selection rules: TTY vs `CI=true` vs explicit flag — which wins?
- Relationship to the existing `--verbosity` levels (Quiet/Minimal/Normal/Verbose). Are modes orthogonal to verbosity, or do modes preset verbosity?
- Tool wrappers (`DotNetBuild` et al.): capture-and-replay-on-failure for `ai` mode, or always stream?
- Schema for `json` mode: stable? versioned? what's the minimum useful payload for an agent?
- Does this become the canonical first-party `IOutputSink` implementation that lands with [RFC #2](https://github.com/ChrisonSimtian/Fallout/issues/98)?

ChrisonSimtian:
Cross-reference: the [AI-native build interface](https://github.com/Fallout-build/Fallout/milestone/24) milestone (#648) picks up the surrounding pieces. The `json` mode schema question raised here is answered in part by #643 (durable run summary and structured diagnostics), and the build-model side is #642 (`describe --json`). This issue stays the home for the output-mode design and default-selection rules.

PR #565, opened 2026-07-26
Add the Information verbosity rung and map levels explicitly

### LINKED ISSUE #556: Add the missing Information verbosity rung between Normal and Minimal
### Problem
There is no verbosity setting that shows information-and-above. `Verbosity` maps onto Serilog minimum levels like this ([src/Fallout.Build/Logging.cs:33-50](https://github.com/Fallout-build/Fallout/blob/main/src/Fallout.Build/Logging.cs#L33-L50)):

| `--verbosity` | Serilog minimum | Effect |
|---|---|---|
| `Verbose` | `Verbose` | everything |
| `Normal` (default) | **`Debug`** | all external tool stdout |
| `Minimal` | `Warning` | drops `[INF]`, including the `> tool args` invocation lines |
| `Quiet` | `Error` | errors only |

External tool stdout defaults to `LogEventLevel.Debug` ([src/Fallout.Tooling/ToolTasks.Logger.cs:42](https://github.com/Fallout-build/Fallout/blob/main/src/Fallout.Tooling/ToolTasks.Logger.cs#L42)), so the default verbosity streams every line of `dotnet`, `git` and `npm` output as `[DBG]`. The only way down is `Minimal`, which also hides which commands ran. There is no "show me the build steps, not their stdout" setting.

Two supporting defects in the same mapping:

- `Verbosity` and `LogLevel` are converted by raw ordinal cast — `(Verbosity)Logging.Level` ([src/Fallout.Build/FalloutBuild.Statics.cs:69-70](https://github.com/Fallout-build/Fallout/blob/main/src/Fallout.Build/FalloutBuild.Statics.cs#L69-L70)). Two unrelated enums coupled by declaration order, so adding a member to either silently re-maps the other.
- The `Logging.Level` getter throws `ArgumentOutOfRangeException` for the `Information` and `Fatal` minimums. `LoggingLevelSwitch` defaults to `Information`, so any read before `FalloutBuild`s static constructor runs throws.

### Outcome
A verbosity setting that shows `[INF]` and above — build steps and tool invocations, without tool stdout. `Verbosity` and `LogLevel` cover the standard `ILogger`/Serilog level set, and converting between them is explicit.

### Acceptance criteria
- [ ] A verbosity setting exists whose Serilog minimum is `Information`, and `--verbosity <that value>` shows `[INF]`/`[WRN]`/`[ERR]` while suppressing `[DBG]` tool stdout
- [ ] `Verbosity` <-> `LogLevel` conversion is an explicit mapping, not an ordinal cast
- [ ] `Logging.Level` returns a value for every `LogEventLevel` instead of throwing
- [ ] Round-trip specs cover every `Verbosity` and every `LogEventLevel`
- [ ] Existing spellings of `--verbosity` keep working and no existing enum member changes numeric value

### Notes
Deliberately non-breaking: members are added, not renumbered. Whether the *default* verbosity moves off the `Debug` firehose is a follow-up decision — this issue is about making the rung exist. Related: #104 (terse default output), #391 (output architecture).

ChrisonSimtian:
Adds the missing rung between `Minimal` and `Normal`. There was no verbosity setting that showed information and above.

### What changed
- Append `Verbosity.Information`, `LogLevel.Information` and `LogLevel.Critical`, so both enums cover the same severities as `ILogger` and Serilog. Members are appended, not renumbered, so existing numeric values are unchanged.
- New `LevelMapping` holds exhaustive conversions between `Verbosity`, `LogLevel` and Serilog `LogEventLevel`. This replaces the ordinal casts in `FalloutBuild.Verbosity` and the partial switch statements in `Logging.Level`.
- `Logging.Level` returns a value for every `LogEventLevel`. It used to throw for the Information and Fatal minimums, and `LoggingLevelSwitch` defaults to Information.
- `VerbosityMappingAttribute` takes an `Information` value, so tool wrappers can map the new rung.
- `SchemaUtility` reflects the `Verbosity` members instead of listing them inline.

### Why
External tool output defaults to the Serilog Debug level, and `Verbosity.Normal` selects Debug as its minimum. The default therefore streams every line of `dotnet`, `git` and `npm` output as `[DBG]`. The only quieter setting was `Minimal`, which also hides `[INF]` and so hides which commands ran.

Measured on a build that shells out to `git` and `dotnet`:

| `--verbosity` | `[DBG]` | `[INF]` | `[WRN]` | `[ERR]` |
|---|---|---|---|---|
| `Normal` | 47 | 3 | 2 | 2 |
| `Information` | 0 | 3 | 2 | 2 |
| `Minimal` | 0 | 0 | 2 | 2 |

At `Information` the tool invocation lines survive, so you still see `> /usr/bin/git --version` without the tool output.

Two things a reviewer should weigh:

- **The default is unchanged.** `Normal` still selects Debug. Moving the default off the firehose is a separate decision, and it belongs with #104.
- **`LogLevel.Critical` maps to `Verbosity.Quiet`.** `Verbosity` has no Critical rung, and adding one would give users a setting that hides ordinary errors. This is the one lossy direction, and a spec pins it.

`Verbosity` members are appended rather than ordered by how much they show. That reads oddly in the source, but renumbering a public enum is a breaking change for consumers that compile the constants in. The mapping is explicit precisely so declaration order stops mattering. Specs pin the existing numeric values.

### What the snapshot churn is
The added enum member flows into generated output: four `SchemaUtilitySpecs` schemas, both `CompletionUtilitySpecs` item lists, the TeamCity generated configuration, and the hardcoded value set in `ParameterServiceSpecs`. All updated.

### Verification
`dotnet test` on `Fallout.Build.Specs`, `Fallout.Common.Specs`, `Fallout.Cli.Specs` and `Fallout.Tooling.Specs`. 27 new specs pass, 8 of which failed before the mapping commit. Behaviour confirmed end to end against a build that shells out to real tools, per the table above.

Closes #556

### DIFF
--- a/src/Fallout.Build/LevelMapping.cs
+++ b/src/Fallout.Build/LevelMapping.cs
@@ -1,0 +1,66 @@
+using System;
+using System.Linq;
+using Serilog.Events;
+
+namespace Fallout.Common;
+
+/// <summary>
+/// Explicit conversions between <see cref="Verbosity"/>, <see cref="LogLevel"/> and Serilog's
+/// <see cref="LogEventLevel"/>.
+/// </summary>
+/// <remarks>
+/// These used to be ordinal casts, which coupled three independently-owned enums by declaration
+/// order: adding a member to any of them silently re-mapped the others. Keep the conversions
+/// exhaustive and explicit. See #556.
+/// </remarks>
+internal static class LevelMapping
+{
+    public static LogLevel ToLogLevel(this Verbosity verbosity)
+        => verbosity switch
+        {
+            Verbosity.Verbose => LogLevel.Trace,
+            Verbosity.Normal => LogLevel.Normal,
+            Verbosity.Information => LogLevel.Information,
+            Verbosity.Minimal => LogLevel.Warning,
+            Verbosity.Quiet => LogLevel.Error,
+            _ => throw new ArgumentOutOfRangeException(nameof(verbosity), verbosity, message: null)
+        };
+
+    public static Verbosity ToVerbosity(this LogLevel level)
+        => level switch
+        {
+            LogLevel.Trace => Verbosity.Verbose,
+            LogLevel.Normal => Verbosity.Normal,
+            LogLevel.Information => Verbosity.Information,
+            LogLevel.Warning => Verbosity.Minimal,
+            LogLevel.Error => Verbosity.Quiet,
+            // Verbosity has no Critical rung, and adding one would give users a setting that hides
+            // ordinary errors. Quiet is the closest thing a user can ask for.
+            LogLevel.Critical => Verbosity.Quiet,
+            _ => throw new ArgumentOutOfRangeException(nameof(level), level, message: null)
+        };
+
+    public static LogEventLevel ToLogEventLevel(this LogLevel level)
+        => level switch
+        {
+            LogLevel.Trace => LogEventLevel.Verbose,
+            LogLevel.Normal => LogEventLevel.Debug,
+            LogLevel.Information => LogEventLevel.Information,
+            LogLevel.Warning => LogEventLevel.Warning,
+            LogLevel.Error => LogEventLevel.Error,
+            LogLevel.Critical => LogEventLevel.Fatal,
+            _ => throw new ArgumentOutOfRangeException(nameof(level), level, message: null)
+        };
+
+    public static LogLevel ToLogLevel(this LogEventLevel level)
+        => level switch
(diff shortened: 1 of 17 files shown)


src/Fallout.Build/LevelMapping.cs
ChrisonSimtian, line 1: well done claude you've outshined yourself here! Finally understanding certain patterns :-)

dennisdoomen (approved):

Issue #391, opened 2026-06-17
Spectre sink: Logging & output architecture — Serilog event core + Spectre.Console frame + NDJSON machine path


ChrisonSimtian:
## Summary
Replace the inherited NUKE console output with a two-layer model:
- **Serilog** owns the structured event stream (one set of log calls, fanned
  out to a human sink and a machine sink).
- **Spectre.Console** owns the rich frame (run header, end-of-run summary
  table), driven directly by the build executor — NOT routed through a sink.

Adds first-class NDJSON output so agents/tools can shell out to Fallout and
parse a clean machine stream.

## Motivation
Supports the AI-native CLI goals: NDJSON streaming, `describe --json`,
structured errors, semantic exit codes. The structured-event model maps a
single set of log calls onto both a pretty console and a machine stream, so
human and agent consumers stay in sync from one source of truth.

## Design notes
Two layers, kept strictly separate. The trap is driving a Spectre
`Progress`/`Live` region from inside an `ILogEventSink` — those are stateful
console-region takeovers and can't be fed by decoupled events. Logging layer
= the event stream; presentation layer = the frame.

## Tasks
- [ ] `SpectreConsoleSink : ILogEventSink` — level→colour mapping,
      `Markup.Escape` on rendered text (paths/arrays/GUIDs contain `[ ]`),
      `AnsiConsole.WriteException` for exceptions
- [ ] NDJSON sink via `Serilog.Formatting.Compact.CompactJsonFormatter` (CLEF)
      → stdout
- [ ] `OutputMode` flag + logger build that branches: human → Spectre sink,
      json → CLEF sink (branch, don't run both — avoids ANSI corrupting the
      NDJSON stream by construction)
- [ ] stdout/stderr discipline in json mode: NDJSON on stdout, all
      diagnostics/human chatter on stderr
- [ ] Promote end-of-run summary to a Spectre `Table` (executor-driven,
      human mode only)
- [ ] JSON-mode structured summary record (single event with results as
      properties) as the equivalent of the table
- [ ] Wire into the host output/logger configuration (swap inherited console
      sink)

## Acceptance criteria
- `fallout build` renders rich human output incl. summary table; no markup
  parse errors on real-world messages (paths, arrays)
- `fallout build --output json` emits valid NDJSON on stdout, one object per
  line, parseable by `jq`, zero ANSI contamination; diagnostics on stderr
- Exceptions render via Spectre in human mode and are captured (`@x`) in json mode

[...]

PR #571, opened 2026-07-26
Label console lines with the full level name

### LINKED ISSUE #570: Print full level names in console output instead of three-letter abbreviations
### Problem
Console output labels every line with a three-letter level abbreviation ([src/Fallout.Build/Logging.cs:27-28](https://github.com/Fallout-build/Fallout/blob/main/src/Fallout.Build/Logging.cs#L27-L28)):

```
22:49:44 [DBG] git version 2.50.1 (Apple Git-155)
22:49:44 [INF] > /usr/bin/git --version
22:49:44 [WRN] a warning line
22:49:44 [ERR] an error line
```

`VRB`, `DBG`, `INF`, `WRN`, `ERR` and `FTL` are not self-explaining. A reader has to learn the codes, and `VRB` versus `DBG` is not obvious even once you have.

The abbreviations exist to keep the level column a fixed width so messages line up. That is worth having, but it is not worth paying for with abbreviations — alignment should come from the layout, not from truncating the words.

### Outcome
Console lines are labelled with the full level name.

```
22:49:44 [Debug] git version 2.50.1 (Apple Git-155)
22:49:44 [Information] > /usr/bin/git --version
22:49:44 [Warning] a warning line
22:49:44 [Error] an error line
```

### Acceptance criteria
- [ ] Console output shows `[Debug]`, `[Information]`, `[Warning]`, `[Error]`, `[Verbose]`, `[Fatal]`
- [ ] The errors-and-warnings summary block uses the same names
- [ ] Level colouring is unaffected
- [ ] Specs cover every level for both console templates

### Notes
Message alignment is deliberately not preserved here. Column layout is #391s job, where Spectre owns the frame and can indent or lay out a real column instead. Scope is the console templates only: the `build.log` file templates keep their compact single-letter column format.

ChrisonSimtian:
Replaces the three-letter level codes in console output with the level names. `VRB`, `DBG`, `INF`, `WRN`, `ERR` and `FTL` are not self-explaining.

### What changed
- `{Level:u3}` becomes `{Level}` in the two console output templates.
- Add `OutputTemplateLevelSpecs`, covering every level for both templates through the same formatter the console sink uses.

Before and after, same build:

```
22:49:44 [DBG] git version 2.50.1        01:43:13 [Debug] git version 2.50.1
22:49:44 [INF] > /usr/bin/git --version  01:43:13 [Information] > /usr/bin/git --version
22:49:44 [WRN] a warning line            01:43:13 [Warning] a warning line
22:49:44 [ERR] an error line             01:43:13 [Error] an error line
```

The errors-and-warnings summary block picks it up too, since it shares a template.

### Why
The codes existed to keep the level column a fixed width so the messages lined up. **That alignment is deliberately not preserved here.** Laying out a real column belongs to #391, where Spectre owns the frame and can indent properly rather than truncating the words. Padding the level to a fixed width is a one-token change (`{Level,-11}`) if a reviewer would rather have ragged-free columns in the meantime.

Two notes on scope:

- **The `build.log` file templates are unchanged.** They use a compact `{Level:u1}` single-letter column in a pipe-delimited, fixed-width layout. That format is for scanning and diffing a log file, not for reading along with a build.
- **Nothing depended on the abbreviations** — no Verify snapshots, no docs, no CI annotation path, no log parsing. Level colouring is applied to the `Level` token by the console sink regardless of format, so it is unaffected.

### Verification
`dotnet test` on `Fallout.Build.Specs` and `Fallout.Common.Specs`. 21 new specs pass, 19 of which failed before the change. Confirmed against a real build run, including the errors-and-warnings block.

Closes #570

### DIFF
--- a/src/Fallout.Build/Logging.cs
+++ b/src/Fallout.Build/Logging.cs
@@ -24,8 +24,12 @@ public static class Logging
         ? AnsiConsoleHostTheme.Default256AnsiColorTheme
         : SystemConsoleHostTheme.DefaultSystemColorTheme;
 
-    internal static string ErrorsAndWarningsOutputTemplate => "[{Level:u3}] {ExecutingTarget}: {Message:l}{NewLine}";
-    internal static string StandardOutputTemplate => "[{Level:u3}] {Message:l}{NewLine}{Exception}";
+    // Full level names, not the {Level:u3} codes. Abbreviating bought a fixed-width level column, but
+    // DBG/VRB/FTL are not self-explaining. Aligning the message column is #391's job — Spectre owns
+    // the frame there and can indent properly instead of truncating the words. The build.log
+    // templates below deliberately keep their compact single-letter column.
+    internal static string ErrorsAndWarningsOutputTemplate => "[{Level}] {ExecutingTarget}: {Message:l}{NewLine}";
+    internal static string StandardOutputTemplate => "[{Level}] {Message:l}{NewLine}{Exception}";
     internal static string TimestampOutputTemplate => $"{{Timestamp:HH:mm:ss}} {StandardOutputTemplate}";
 
     private const int TargetNameLength = 20;
--- a/tests/Fallout.Build.Specs/OutputTemplateLevelSpecs.cs
+++ b/tests/Fallout.Build.Specs/OutputTemplateLevelSpecs.cs
@@ -1,0 +1,102 @@
+using System;
+using System.IO;
+using System.Linq;
+using Fallout.Common.Execution;
+using FluentAssertions;
+using Serilog.Events;
+using Serilog.Formatting.Display;
+using Serilog.Parsing;
+using Xunit;
+
+namespace Fallout.Common.Specs;
+
+/// <summary>
+/// Covers the level label in the console output templates. See #570. The templates used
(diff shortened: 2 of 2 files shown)


ChrisonSimtian:
this one im not sure about yet, I could imagine a lot of people would be pissed off because they love it when things are aligned. But then I'd argue we could just do [I] [W] [E] [D] [T] etc.

IT-VBFK:
Hmm.. I would definitely vote for the "one-letter"-thing, despite I don't get what "T" stands for?

ChrisonSimtian:
> Hmm.. I would definitely vote for the "one-letter"-thing, despite I don't get what "T" stands for?

Trace
Log / Verbosity levels are usually (free after `ILogger`):
- [E]xception or [E]rror
- [W]arning
- [I]nformation
- [D]ebug
- [T]race

This is part of a bigger refactoring around introducing `ILogger` as an abstraction layer and making use of `Serilog` (which is already being used)
At some point, at least thats my vision, the output is purely cosmetic sugar and YOU can bring your own renderer (or override the stock standard one to fit your flavour)
The log output itself ideally comes as NDJson, which can be picked up by heaps of other tooling like AI or all sorts of CI runners.

PR #572, opened 2026-07-26
Test BumpDotNetVersionStep through its public ExecuteAsync method

dennisdoomen:
## Why
`BumpDotNetVersionStepSpecs` tested `BumpTargetFramework`/`BumpSdkVersion` directly, which forced those helpers to stay `public` even though only `IMigrationStep.ExecuteAsync` is the step's real contract.

## What
- Make `BumpTargetFramework` private.
- Rewrite the specs to drive the step through `ExecuteAsync` against real files on disk (constructor + `IDisposable.Dispose()` pattern, `AbsolutePath.Temp`), asserting on file content and `Summary` counts.
- Document the constructor/`Dispose()` fixture pattern and AAA-comment convention in `docs/agents/conventions.md`.

Demonstrates testing `IMigrationStep` implementations solely through their public method.


docs/agents/conventions.md
ChrisonSimtian, line 1: nice I like the AAA

ChrisonSimtian (approved):

Issue #573, opened 2026-07-26
Fold ScriptRewriter into RewriteBootstrapScriptsStep


ChrisonSimtian:
### Problem
`src/Fallout.Migrate/Steps/` mixes two shapes. Most files are migration steps that implement `IMigrationStep`. `ScriptRewriter` is a free-floating `static` class that is not a step — it only holds the rewrite rules for one step.

That split was the original design in #507, but it has since been retired one file at a time:

- `CsprojRewriter` was folded into `RewriteCsprojsStep` in `aef6c073`, on the grounds that it had a single caller.
- `CodeRewriter` was folded into `RewriteCsFilesStep` in #528, for the same reason.

`ScriptRewriter` is the last one. It has exactly one caller, [RewriteBootstrapScriptsStep.cs:26](https://github.com/Fallout-build/Fallout/blob/main/src/Fallout.Migrate/Steps/RewriteBootstrapScriptsStep.cs#L26), so the same reasoning applies. Until it moves, `Steps/` teaches two contradictory patterns, and a contributor adding a bootstrap-script rewrite rule has to guess which one to follow.

### Outcome
Every file in `src/Fallout.Migrate/Steps/` is a migration step, and each step owns the rewrite rules for its own artifact type.

### Acceptance criteria
- [ ] `ScriptRewriter`s regex fields and `Rewrite` method live in `RewriteBootstrapScriptsStep`
- [ ] `src/Fallout.Migrate/Steps/ScriptRewriter.cs` is deleted
- [ ] `ScriptRewriterSpecs` is renamed to `RewriteBootstrapScriptsStepSpecs`, file and class
- [ ] `dotnet test tests/Fallout.Migrate.Specs` passes with no test changes beyond the rename
- [ ] No behaviour change

### Notes
Good first issue: a mechanical move of about 50 lines with no design decisions. Copy the shape of `aef6c073` (`Fold CsprojRewriter into RewriteCsprojsStep`) — it is the same change on a sibling step. The convention is written up in #574.

ITaluone:
Ou.. that went along too in #577

ChrisonSimtian:
> Ou.. that went along too in [#577](https://github.com/Fallout-build/Fallout/pull/577)

nice :-) all yours ;-)

PR #574, opened 2026-07-26
Document the fallout-migrate step pattern for agents

ChrisonSimtian:
Writes down the `fallout-migrate` step convention, which currently exists only in commit history.

### What changed
- New "Migration step recipe" section in `docs/agents/conventions.md`, next to the existing tool wrapper recipe.
- A pointer bullet in `AGENTS.md`. That file links `conventions.md` rather than importing it, so the rule needs a line where an agent will actually see it.

### Why
`src/Fallout.Migrate/Steps/` teaches two contradictory patterns. Most files are steps implementing `IMigrationStep`. `ScriptRewriter` is a free-floating static class that only holds the rules for one step. #507 introduced that split, then `aef6c073` began retiring it by folding `CsprojRewriter` into `RewriteCsprojsStep` because it had a single caller.

The convention was never written down, so it came up in review on #528: a new rename rule went into a rewriter class instead of the step. The rule now has a home that does not require commit archaeology.

The recipe states the part that is easy to get backwards. A new rename goes into the step that already covers that file type. A new step is only for a new set of files or a different kind of operation. It also records why a step per rename is wrong: a second pass over the same files, each file written twice, inflated `Summary` edit counts, and rule ordering becoming a hidden dependency between entries in `Migration.steps` instead of plain statement order.

Docs only, no code change.

The remaining leftover is tracked in #573.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -79,5 +79,6 @@ Full conventions + what-not-to-do list: [docs/agents/conventions.md](docs/agents
 
 - The `build/Build.*.cs` files are the canonical example of how to consume the framework — read these when reasoning about user-facing APIs.
 - `src/Fallout.Common/Tools/<Tool>/<Tool>.json` files are the source of truth for tool wrappers; the `.cs` next to them is generated.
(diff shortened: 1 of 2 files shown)


ChrisonSimtian:
@dennisdoomen claude tripped over the new migrationstep concept, found two approaches and picked the old one. This should help AI to understand it a bit better and pick the right approach

dennisdoomen (approved):

PR #576, opened 2026-07-26
Switch the global tool during migration, and target the kept package id

ChrisonSimtian:
> **Stacked on #581.** This branch is rebased onto it, so the diff below includes #581's commits until that merges. Review the last commit (`a278be9b`) for this PR's own change, and merge #581 first.

Makes `fallout-migrate` fix the whole tool-id problem, not just the repo half of it.

## Point the migration at the id we kept

#581 settled on keeping `Fallout.GlobalTool` rather than moving to `Fallout.GlobalTools`, so the current id and the retired list swap places:

- **current:** `fallout.globaltool`
- **retired:** `nuke.globaltool`, `fallout.cli`, `fallout.globaltools`

A consumer already pinning `fallout.globaltool` now migrates to nothing. That is the point of keeping the id, and there is a spec for it.

## Switch the machine's global tool

Rewriting `.config/dotnet-tools.json` only fixes the repo. A machine-wide install of a retired tool stays behind, and two packages providing the `fallout` command conflict — so the repo-only fix left the machine broken.

New `SwitchGlobalToolStep` reads `dotnet tool list --global`, uninstalls each retired id that is installed, then installs the current one at the resolved version.

- Runs after `RewriteToolManifestStep`. Repo first, so an interrupted run still leaves the committed state correct.
- Best-effort throughout. No `dotnet` on `PATH`, no global install, or an unreachable feed produces a `Summary` warning naming the command to run by hand. It never fails the migration.
- `--dry-run` prints the commands without running them.
- Does nothing when no retired id is installed, so the common case is silent.

## Process execution

`SwitchGlobalToolStep` shells out through the framework's own `ProcessTasks` (`src/Fallout.Tooling`). **No new package dependency.**

`CliWrap` was trialled here first and rejected — findings, the side-by-side code, and the measured comparison are in #583. Short version: both versions behave identically, the call site is 31 lines instead of 18, and `ProcessTasks` already handles the pipe-buffer deadlock that makes hand-rolled process runners dangerous. The four places `ProcessTasks` is a worse API (no `CancellationToken`, no async overload, pre-quoted argument string, Serilog by default) are acceptance criteria on #583 rather than a reason to take the package.

Referencing `Fallout.Tooling` does pull `Serilog` and `NuGet.Packaging` into the migrate tool — one direct dependency traded for two transitive ones. Called out because `Fallout.Migrate` is otherwise deliberately lean.

## Also

- `.config/dotnet-tools.json` moves to `fallout.globaltool` `10.3.49`. The dogfood pin could not move while the id was `Fallout.GlobalTools`, because nothing above `10.4.0-rc.4` was published there. Verified `dotnet tool restore` resolves it.
- `docs/migration/tool-package-id.md` rewritten around the kept id. It is now a "move off a retired id" page, and most readers need no action.

## Verification

- `Fallout.Migrate.Specs` 63/63, including 8 new specs for the tool-list parsing and install-command rendering
- A direct probe through `ProcessTasks` returns the same seven global tool ids the CliWrap version did
- `dotnet build fallout.slnx` clean
- `fallout-migrate --dry-run` against a scratch repo pinning `fallout.cli` rewrites the manifest as expected
- `dotnet tool restore` resolves the new dogfood pin

## Notes

- The step's process calls are not unit-tested — that would install software on the machine running the specs. The parsing and command rendering are covered; the `dotnet tool` calls themselves are not.
- #582 tracks the durable rule that stops the id moving again.


src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
ChrisonSimtian, line 1: ok something here went wrong @claude pretty sure you went a bit too far here

ChrisonSimtian:
alternatively we rename this back to singular

src/Fallout.Migrate/Steps/ResolveFalloutVersionStep.cs
```diff
@@ -40,10 +40,11 @@ internal sealed class ResolveFalloutVersionStep : IMigrationStep
...
 
     /// <summary>
-    /// NuGet's flat-container index for <see cref="PackageId"/>: a JSON document listing every
-    /// published version (stable and prerelease), oldest first.
+    /// The dotnet-tool package id pinned in a migrated <c>dotnet-tools.json</c>. Resolved separately
+    /// from <see cref="PackageId"/>: the tool id was introduced later, so it has its own, much
+    /// shorter, set of published versions (#575).
     /// </summary>
-    private static readonly Uri flatContainerIndex = new($"https://api.nuget.org/v3-flatcontainer/{PackageId}/index.json");
+    private const string ToolPackageId = RewriteToolManifestStep.CurrentToolId;
```
ChrisonSimtian, line 47: @claude same as in #581 you can point this right at the `PackageId` property of the Fallout Cli project, instead of hardcoding a string value here. This way we always resolve the correct package!

src/Fallout.Migrate/Steps/RewriteToolManifestStep.cs
```diff
@@ -0,0 +1,125 @@
...
+{
+    /// <summary>The current published tool package id.</summary>
+    internal const string CurrentToolId = "fallout.globaltool";
+
+    /// <summary>Tool ids this step migrates away from.</summary>
+    internal static readonly IReadOnlyList<string> RetiredToolIds =
+    [
+        "nuke.globaltool",
+        "fallout.globaltools",
+        "fallout.cli"
```
ChrisonSimtian, line 32: there was a brief moment in time where this package did exist. Lets leave this in here just in case

ChrisonSimtian:
alright, this got some smarts now:
It ALWAYS runs the migration, it has a list of retired variants that we replace and it goes and makes sure the user has the CLI tool installed as per the Cli-Project's `PackageId`. This means should we change the PackageID in the future, the migration will automatically(!!) pick it up and migrate it for the user

src/Fallout.Migrate/Steps/SwitchGlobalToolStep.cs
```diff
@@ -0,0 +1,280 @@
...
+        Execute(context, summary);
+
+        return Task.CompletedTask;
+    }
+
+    /// <summary>
+    /// The step's actual work. Synchronous because <see cref="ProcessTasks"/> is; the interface's
+    /// async signature is satisfied by the wrapper above.
+    /// </summary>
+    private static void Execute(MigrationContext context, Summary summary)
```
ITaluone: What is the big advantage of "just" forwarding to a private method with the exact same parameters?

ChrisonSimtian: thats a @claude special. Go forth AI and fix your own slop

PR #577, opened 2026-07-27
Refactor migration specs

### LINKED ISSUE #573: Fold ScriptRewriter into RewriteBootstrapScriptsStep

ITaluone:
Refactors all `Fallout.Migrate.Specs` step tests to exercise the full `Migration` pipeline instead of calling steps directly, and applies the AV1600 behavior-focused naming convention.

### What changed
- `RewriteCsprojsStepSpecs`, `RewriteCsFilesStepSpecs`, `BumpDotNetVersionStepSpecs`, `ScriptRewriterSpecs` now write real files to a temp directory, run them through `Migration.RunAsync()`, and assert on the file content read back, instead of calling `RewriteCsprojsStep`/`RewriteCsFilesStep`/`BumpDotNetVersionStep`/`ScriptRewriter` directly.
- Renamed all `[Fact]`/`[Theory]` methods in `RewriteCsFilesStepSpecs`, `BumpDotNetVersionStepSpecs`, and `ScriptRewriterSpecs` to short, present-tense, snake_case descriptions of observable behavior, matching [docs/agents/conventions.md](docs/agents/conventions.md)'s AV1600 convention already used elsewhere.

### Why?
Calling steps in isolation let the specs drift from what the real migration pipeline produces (e.g. hardcoded version strings that don't match `ResolveFalloutVersionStep`'s live resolution.

### Verification
- Traced each test's input through the full pipeline manually to confirm expected edit counts/content.
- All Migration.Specs green

### Additional things to mention:
Does not refactor the `BumpDotNetVersionStepSpecs` in favor of #572 

From: https://github.com/Fallout-build/Fallout/pull/519#issuecomment-5085001779


src/Fallout.Migrate/Steps/RewriteBootstrapScriptsStep.cs
```diff
@@ -1,15 +1,35 @@
...
 using System.Threading.Tasks;
 using Fallout.Common.IO;
 using Fallout.Migrate.Common;
 
 namespace Fallout.Migrate.Steps;
 
 /// <summary>
-/// Rewrites <c>build.cmd</c>, <c>build.ps1</c>, and <c>build.sh</c> at the repository root, when
-/// present, via <see cref="ScriptRewriter"/>.
+/// Rewrites bootstrap scripts (<c>build.cmd</c>/<c>build.ps1</c>/<c>build.sh</c>): <c>dotnet nuke</c>
```
ChrisonSimtian, line 9:
```suggestion
/// Rewrites bootstrap scripts (<c>build.cmd</c>/<c>build.ps1</c>/<c>build.sh</c>): <c>dotnet fallout</c>
```
right? or am I a bit lost here? :-)

ChrisonSimtian (approved):
lgtm, iirc then there is a `--dryrun` flag in the migrator. Should probably support this 

well we should probably enforce a dryrun method to be implemented in the `IMigrationStep` construct and do some architecture extra work to make sure all steps work like that, but lets not go crazy in this first step :-) Just thinking out loud :D

src/Fallout.Migrate/Steps/RewriteBootstrapScriptsStep.cs
ITaluone: Lost ;)

It tells what this changes: from `dotnet nuke` to `...their Fallout equivalents...`

PR #580, opened 2026-07-27
Fix the user facing migrate command description

ITaluone:
Fix wording in `fallout-migrate`'s user-facing `--help` description.

### What changed
- Corrects the `[Description]` text on `MigrateCommand` (`src/Fallout.Migrate/MigrateCommand.cs:16-29`) — the summary users see when running `fallout-migrate --help`.

### Why
The description is the first thing consumers of the migration tool read; it needs to accurately describe what the migration does.

### Verification
- ` dotnet run .\Fallout.Migrate.csproj -- --help` shows the fixed help text

### DIFF
--- a/src/Fallout.Migrate/MigrateCommand.cs
+++ b/src/Fallout.Migrate/MigrateCommand.cs
@@ -24,6 +24,7 @@ namespace Fallout.Migrate;
                - Rewrites `dotnet nuke` -> `dotnet fallout` and legacy NUKE_* env vars
                  in build.cmd / build.ps1 / build.sh
                - Renames .nuke/ to .fallout/
+               - Removes legacy NUKE_ENTERPRISE_TOKEN check from bootstrapper scripts, if exists
                - Prints a summary of files changed and warnings to address manually
              """)]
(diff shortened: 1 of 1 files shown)


src/Fallout.Migrate/MigrateCommand.cs
```diff
@@ -40,10 +41,10 @@ public override async Task<int> ExecuteAsync(CommandContext context, MigrateSett
         var rootDirectory = ResolveRootDirectory(settings.Path);
         if (rootDirectory is null)
         {
-            Console.Error.WriteLine(
+            await Console.Error.WriteLineAsync(
```
dennisdoomen: 🤔 Why did you need these changes?

ITaluone: "need" is relative :)

Rider gave me the "has an async overload" suggestion..

reverted this

~~BTW.. I don't get why Rider keeps suggesting this, given, that the .editorconfig rule disables this~~ 🤔
Ou.. I see.. this only counts for tests..

dennisdoomen (approved):

PR #581, opened 2026-07-27
Keep publishing the CLI as Fallout.GlobalTool

ChrisonSimtian:
Keep publishing the dotnet tool as **`Fallout.GlobalTool`** (singular) instead of the `Fallout.GlobalTools` id introduced in f0208100.

`Fallout.GlobalTool` has 52 published versions on nuget.org (10.2.12 → 10.3.49) — that's the id the entire existing install base already has pinned. Staying on it means **no consumer migrates, no id gets unlisted, and no migration guide is needed**. Assembly/namespace stay `Fallout.Cli`; the command stays `fallout`.

The plural id shipped exactly one prerelease, `Fallout.GlobalTools 10.4.0-rc.4`, so the blast radius is rc testers only — not a GA break, hence no `breaking-change` label. The README note covers them.

## Changes

- **`Fallout.Cli.csproj`** — `<PackageId>Fallout.GlobalTool</PackageId>`, with a comment recording why the package id deliberately doesn't match the assembly name.
- **`templates/dotnet-tools.json` + `BuildScaffolder`** — `fallout :setup` writes a `fallout.globaltool` pin into consumer repos again.
- **`UpdateNotificationAttribute`** — in-tool prompt says `dotnet tool update Fallout.GlobalTool -g`.
- **`build/Build.cs`** — drive-by fix: the `Install` target installed `Solution.Fallout_Cli.Name` (`"Fallout.Cli"`), which never matched the packed nupkg id and has been broken since the id was first decoupled. Now uses a `GlobalToolPackageId` constant.
- **Docs** — repo-wide sweep of the plural id: README, CONTRIBUTING, introduction, 01-installation, 02-setup, 03-execution, github-actions, rebrand-plan. Deleted `docs/migration/from-globaltool-to-cli.md` (documented a `GlobalTool → Cli` rename that is now explicitly never happening; already `draft: true`, no inbound links).

## Deliberately not in this PR

`.config/dotnet-tools.json` still pins `fallout.cli 10.3.45` — the repo's own dogfood pin. Flipping it needs a published `Fallout.GlobalTool` newer than 10.3.49; changing it here would regress `dotnet tool restore` for the bootstrappers.

## Verification

- `dotnet pack src/Fallout.Cli` → `Fallout.GlobalTool.10.0.0-preview.55.g748120854f.nupkg`
- `Fallout.Cli.Specs` 47/47 passing; `build/_build.csproj` compiles

## Follow-ups

- Cherry-pick onto `release/v10.4` before 10.4.0 GA, so GA continues the `Fallout.GlobalTool` sequence from 10.3.49.
- Deprecate `Fallout.GlobalTools 10.4.0-rc.4` on nuget.org with `Fallout.GlobalTool` as the alternate package.
- Supersedes most of #576 — that PR moves the repo *onto* the plural id and adds a `RewriteToolManifestStep` that rewrites consumer manifests to `fallout.globaltools`. Needs rebasing to the singular id or closing.


build/Build.cs
```diff
@@ -283,8 +289,8 @@ IEnumerable<AbsolutePath> NuGetPackageFiles
         .DependsOn<IPack>()
         .Executes(() =>
         {
-            SuppressErrors(() => DotNet($"tool uninstall -g {Solution.Fallout_Cli.Name}"), logWarning: false);
-            DotNet($"tool install -g {Solution.Fallout_Cli.Name} --add-source {OutputDirectory} --version {DefaultDeploymentVersion}");
```
ChrisonSimtian, line 287: @claude we should be able to grab the `PackageId` from this Project. If not then that is a new attribute we want to pass through (mark this as an issue, raise a PR and lets wait for it to be merged before we merge this PR)
This way we dont need to maintain yet another hardcoded string somewhere, which reduces the blast radius of changing this back later on

ChrisonSimtian: Good call — it was already reachable, no new attribute needed.

`Fallout.ProjectModel` has exposed MSBuild property access on the solution model since the NUKE days (`ProjectExtensions.GetProperty`, `src/Fallout.ProjectModel/Project.Properties.cs`), so the object path works as-is:

```csharp
var packageId = Solution.Fallout_Cli.GetProperty("PackageId");   // -> Fallout.GlobalTool
```

Pushed in f4313ca8 — the `GlobalToolPackageId` constant is gone and the csproj is the single source of truth, so changing the id later touches exactly one file.

Worth noting the same MSBuild-evaluation path is already exercised by this build at `Build.cs:98` (`project.GetTargetFrameworks()`), so this is not a new runtime dependency. Verified end-to-end against `fallout.slnx`: `GetProperty("PackageId")` on the `Fallout.Cli` project returns `Fallout.GlobalTool`.

No blocking issue/PR needed — this PR is unblocked from your review's perspective.

build/Build.cs
```diff
@@ -283,8 +283,13 @@ IEnumerable<AbsolutePath> NuGetPackageFiles
...
+            // Read the id off the csproj rather than hardcoding it: the tool's PackageId is
+            // deliberately decoupled from its project/assembly name (Fallout.Cli packs as
+            // Fallout.GlobalTool), so Solution.Fallout_Cli.Name would install the wrong package.
+            var packageId = Solution.Fallout_Cli.GetProperty("PackageId");
```
ChrisonSimtian, line 289: probably deserves a standalone property, but thats a job for another day

ChrisonSimtian:
Follow-ups from this thread are now in place:

- **#582** — issue covering the full rename history and the direction: package id is a consumer contract, project/assembly/namespace are internal and free to move. Includes acceptance criteria for an ADR, an explicit `PackageId` on every packable project, a published-id registry, a build check, and the rule that any future normalisation happens once at a yearly major rather than incrementally.
- **#576** — rebased onto this branch and retargeted at `Fallout.GlobalTool`. It also gains a `SwitchGlobalToolStep` that uninstalls a retired global tool and installs the current one, so a migration fixes the machine as well as the repo.

One useful side effect: with the id back to `Fallout.GlobalTool`, the dogfood pin in `.config/dotnet-tools.json` is no longer blocked. `fallout.globaltool 10.3.49` is published and restores, so #576 flips the pin off `fallout.cli 10.3.45`. That was the one thing this PR deliberately left alone.

src/Fallout.Build/Execution/Extensions/UpdateNotificationAttribute.cs
```diff
@@ -35,7 +35,7 @@ private static void Notify()
             {
                 "--- UPDATE RECOMMENDED FROM 5.1.0 ---",
                 "1. Update your global tool",
-                "   dotnet tool update Fallout.GlobalTools -g",
+                "   dotnet tool update Fallout.GlobalTool -g",
```
ChrisonSimtian, line 38: @claude lets make this magic and grab the packageId from our other project here, ok?
Or if we have to hardcode this, then maybe we should make it a const string and centralise this somewhere in the inner core instead.

ChrisonSimtian: Filed as **#584**. Short answer: the magic works in `Build.cs` but cannot work here, and the reason is worth recording.

`Solution.Fallout_Cli.GetProperty("PackageId")` works in `build/Build.cs` because that evaluates *this repository's* csproj at build-authoring time. `UpdateNotificationAttribute` ships inside `Fallout.Common` and runs in a **consumer's** build, where no Fallout csproj exists to read. There is nothing to walk the object path to.

So it is either a hardcoded string, or the id gets flowed in at compile time — one MSBuild property in a shared `.props` feeding both `<PackageId>` in `Fallout.Cli.csproj` and a generated constant. That second option is the real "grab it from the other project", and it is what #584 asks for.

Leaving the literal in place on this PR, per your call that centralising is its own PR.

One finding from the scan that changes the shape of that PR: **`Fallout.Core` cannot be the home as it stands.** Core targets `netstandard2.1;net10.0`. The existing `Constants.cs` lives in `Fallout.Build.Shared` (`netstandard2.0;net10.0`), and `Fallout.SourceGenerators` consumes `Fallout.Build.Shared.dll` directly (`Fallout.SourceGenerators.csproj:42`) — Roslyn requires `netstandard2.0`. Moving these constants into Core today would cut the generator off from them. #584 carries that as an open question with three ways out.

You were also right that there is more of this than the one string. The scan found ~11 settings scattered as call-site consts, including four values declared twice: the `Fallout.Common` package id, the nuget.org v3 index URL, the flat-container URL, and the tool id. Full table in #584.

docs/rebrand-plan.md
```diff
@@ -45,7 +45,7 @@ This is locked by the bridge-package design — `[TypeForwardedTo]` requires the
 | `Nuke.Common.Utilities` (+ `.Collections`, `.Net`) | `Fallout.Common.Utilities` | |
 | `Nuke.Common.ValueInjection` | `Fallout.Common.ValueInjection` | |
 | `Nuke.Components` | `Fallout.Components` | |
-| `Nuke.GlobalTool` (+ `.Rewriting.Cake`) | `Fallout.Cli` | Project/namespace/assembly is `Fallout.Cli`. The **published package id is `Fallout.GlobalTools`** (`dotnet tool install Fallout.GlobalTools`) — decoupled from the assembly name on purpose. Command name stays `fallout`. |
+| `Nuke.GlobalTool` (+ `.Rewriting.Cake`) | `Fallout.Cli` | Project/namespace/assembly is `Fallout.Cli`. The **published package id stays `Fallout.GlobalTool`** (`dotnet tool install Fallout.GlobalTool`) — deliberately decoupled from the assembly name and deliberately *not* renamed, so the existing `10.x` install base never has to migrate. Command name stays `fallout`. |
```
dennisdoomen, line 48: 🤔 Do we still care about this file?

src/Fallout.Build/Execution/Extensions/UpdateNotificationAttribute.cs
```diff
@@ -35,7 +35,7 @@ private static void Notify()
             {
                 "--- UPDATE RECOMMENDED FROM 5.1.0 ---",
```
dennisdoomen: 🤔5.1.0? Were does that come from?

dennisdoomen (approved):
🤔 Don't you want this on 10.4? If so, it should target the release branch.

docs/rebrand-plan.md
ChrisonSimtian: no not really, it just pollutes AI into thinking it needs to maintain it.
@claude lets raise a PR to remove this file (it'll be in the git history forever)

src/Fallout.Build/Execution/Extensions/UpdateNotificationAttribute.cs
ChrisonSimtian: yeah I saw that too... I told claude to look at the history, it looked into NUKEs history as well and probably looked a bit too far into the past :D

ChrisonSimtian:
```suggestion
                "",
```

build/Build.cs
ChrisonSimtian: Filed as #589 — a typed `PackageId` accessor on `ProjectExtensions`, sibling of the `GetTargetFrameworks()` / `GetOutputType()` accessors already in `Project.Misc.cs`. It carries the MSBuild fallback rule (explicit `<PackageId>` wins, else `AssemblyName`) so it stays correct for the packable projects that do not set one, and flips this `Install` target over to it. `GetProperty` stays as the escape hatch.

Leaving the string lookup here for now, per "another day".

src/Fallout.Build/Execution/Extensions/UpdateNotificationAttribute.cs
ChrisonSimtian: Traced it: hardcoded in NUKE `9885956b` ("Reimplement backwards-compatibility", 2021-03-08), when 5.1.0 was the current NUKE version — and never touched in the five years since. So it named a version no Fallout consumer has ever run.

Fixed in a4561062.

ChrisonSimtian: Took the intent, kept a header: `--- UPDATE RECOMMENDED ---` (a4561062).

`[UpdateNotification(Priority = 10)]` sits on `FalloutBuild` itself, so this is live for every consumer — it fires through `Host.Warning` at the end of a server build in any repo still on the pre-`.fallout` layout. A bare `""` would leave "1. Update your global tool" landing with nothing saying why. Say the word if you would rather it be silent about that.

src/Fallout.Build/Execution/Extensions/UpdateNotificationAttribute.cs
ChrisonSimtian: Confirmed as-is on this PR; #584 owns the centralisation.

One thing worth carrying over into that work: `Fallout.Core` cannot be the home as it stands. Core targets `netstandard2.1;net10.0`, while `Fallout.SourceGenerators` consumes `Fallout.Build.Shared.dll` directly and Roslyn requires `netstandard2.0` — so lifting these constants into Core cuts the generator off from them.

ChrisonSimtian:
Review round two — all threads answered.

**On 10.4** (@dennisdoomen): yes, it needs to land there, but via `main` first rather than retargeting.

`main` is the integration trunk and `release/YYYY`-style branches are the production line, so anything that is not a release-line-only fix lands on `main` and gets promoted — there is already a `promote/main-to-v10.4` lane doing exactly that. Retargeting would strand the `Install`-target fix and the docs sweep on the release branch and let `main` drift straight back onto the plural id. Cherry-pick before 10.4.0 GA is tracked in the PR description, so GA continues the `Fallout.GlobalTool` sequence from 10.3.49.

**Pushed:**

- a4561062 — dropped the stale `5.1.0` from the update-notification header. It was hardcoded in NUKE `9885956b` (March 2021) and never updated, so it named a version no Fallout consumer has run. Kept a `--- UPDATE RECOMMENDED ---` header rather than the bare `""`, since `[UpdateNotification]` is live on `FalloutBuild` and the numbered steps need context in a server-build warning.
- f7226b47 — the `<PackageId>` comment now points at #582 instead of `docs/rebrand-plan.md`, which is being removed.

**Filed:**

- #589 — typed `PackageId` accessor on `ProjectExtensions`, so `Install` stops naming an MSBuild property in a string. Deferred out of this PR.
- #584 — already open, owns the wider constants centralisation (including the tool id literal in `UpdateNotificationAttribute`, which cannot be resolved from a csproj because it runs in a consumer build).

**Separate PR incoming** for removing `docs/rebrand-plan.md`. It is not a clean delete — five inbound references, one of them shipped to users as the `NukeMigrationAnalyzer` `helpLinkUri` — so it gets its own change rather than riding along here.

docs/rebrand-plan.md
ChrisonSimtian: Raised as #591 (draft).

Not a clean delete — five inbound references, and one of them ships: `NukeMigrationAnalyzer.helpLinkUri` uses this doc as the help link on the migration diagnostic in consumers' IDEs, so a plain `git rm` would have shipped a 404. That now points at `docs/migration/from-nuke.md`. The `Fallout.Migrate.Analyzers` README also deep-linked the `#sunset-timeline` anchor.

Two facts from the doc were worth keeping, so they moved to where they are used rather than dying with it: the strict 1:1 prefix-swap rule (and why `[TypeForwardedTo]` locks it) is now inline in the `AGENTS.md` rebrand-status paragraph, and the `PackageId`-vs-assembly-name reason already lives in `Fallout.Cli.csproj`.

Agreed on the diagnosis, for the record — parts of it were actively wrong. The sunset timeline was anchored to a "first `Fallout.<X>` 11.x release" that never shipped.

PR #597, opened 2026-07-29
Add support for uncompressing `.tar.xz` archives

### LINKED ISSUE #596: Fallout (and also former NUKE) lacks of possibility to uncompress a .tar.xz archive
### Description

GZip, Bzip2, Zip.. all of them have compression extensions.

But Xz (.tar.xz) is missing

### Why?

In general, only to get around qirks like that: https://github.com/fluentassertions/fluentassertions/blob/main/Build/CompressionExtensions.cs
And: why shouldn't we support (at least decompressing) .tar.xz files, too? 

### Usage Example

```csharp
// AbsolutePath dir = RootDirectory / "some" / "subfolder";
// dir.TarXzTo(RootDirectory / "bla.tar.xz");
// dir.CompressTo(RootDirectory / "bla.tar.xz");
// This is not possible...

// and respective:
AbsolutePath archive= RootDirectory / "bla.tar.xz";
archive.UnTarXzTo(RootDirectory / "some" / "subfolder1");
archive.UncompressTo(RootDirectory / "some" / "subfolder2");
```

### Alternative

Making this logic in exactly every project where needed...

Yes

ITaluone:
### What changed
- Add `UnTarXzTo` extension on `AbsolutePath`, using `SharpCompress.Readers.ReaderFactory` to extract `.tar.xz`/`.txz` archives.
- Wire it into `UncompressTo` alongside the existing zip/tar.gz/tar.bz2 branches.
- Compressing *to* `.tar.xz` stays unsupported (`CompressTo` still fails fast with a clear message) — this PR only adds the read path.
- Add the `SharpCompress` package reference to `Fallout.Utilities.IO.Compression`.

### Verification
Manual `Target` which runs this extension

### Drawback
Compressing `.tar.xz` files does not work, since the upstream lib also can't do this.

Closes: #596

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -26,6 +26,7 @@
     <PackageVersion Include="Serilog.Sinks.Console" Version="6.1.1" />
     <PackageVersion Include="Serilog.Sinks.File" Version="7.0.0" />
     <PackageVersion Include="SharpZipLib" Version="1.4.2" />
+    <PackageVersion Include="SharpCompress" Version="0.50.1" />
     <PackageVersion Include="System.ComponentModel.Annotations" Version="5.0.0" />
     <PackageVersion Include="System.Net.Http" Version="4.3.4" />
     <PackageVersion Include="System.Security.Cryptography.Xml" Version="10.0.10" />
--- a/src/Fallout.Utilities.IO.Compression/CompressionExtensions.cs
+++ b/src/Fallout.Utilities.IO.Compression/CompressionExtensions.cs
@@ -8,6 +8,8 @@ using ICSharpCode.SharpZipLib.BZip2;
 using ICSharpCode.SharpZipLib.GZip;
 using ICSharpCode.SharpZipLib.Tar;
 using ICSharpCode.SharpZipLib.Zip;
+using SharpCompress.Common;
+using SharpCompress.Readers;
 using ZipFile = ICSharpCode.SharpZipLib.Zip.ZipFile;
 
 namespace Fallout.Common.IO;
@@ -28,6 +30,10 @@ public static class CompressionExtensions
         {
             directory.TarBZip2To(archiveFile, filter);
         }
+        else if (archiveFile.HasExtension(".tar.xz", ".txz"))
+        {
+            Assert.Fail($"Compressing a .tar.xz archive currently not supported. Archive file: '{Path.GetFileName(archiveFile)}'");
+        }
         else
         {
             Assert.Fail($"Unknown archive extension for archive '{Path.GetFileName(archiveFile)}'");
@@ -47,6 +53,10 @@ public static class CompressionExtensions
         else if (archiveFile.HasExtension(".tar.bz2", ".tbz2", ".tbz"))
         {
             archiveFile.UnTarBZip2To(directory);
+        }
+        else if (archiveFile.HasExtension(".tar.xz", ".txz"))
+        {
+            archiveFile.UnTarXzTo(directory);
         }
         else
         {
@@ -147,6 +157,26 @@ public static class CompressionExtensions
     public static void UnTarBZip2To(this AbsolutePath archiveFile, AbsolutePath directory)
     {
         UncompressTar(archiveFile, directory, x => new BZip2InputStream(x));
+    }
+
+    public static void UnTarXzTo(this AbsolutePath archive, AbsolutePath directory)
+    {
+        using Stream stream = File.OpenRead(archive);
+        using var reader = ReaderFactory.OpenReader(stream);
+
+        while (reader.MoveToNextEntry())
+        {
+            if (reader.Entry.IsDirectory)
+            {
+                continue;
+            }
+
+            reader.WriteEntryToDirectory(directory, new ExtractionOptions
+            {
+                ExtractFullPath = true,
+                Overwrite = true
+            });
+        }
     }
 
     private static void CompressTar(
--- a/src/Fallout.Utilities.IO.Compression/CompressionExtensions.cs
+++ b/src/Fallout.Utilities.IO.Compression/CompressionExtensions.cs
@@ -16,6 +16,19 @@ namespace Fallout.Common.IO;
 
 public static class CompressionExtensions
 {
+    /// <summary>
+    ///     Compresses <paramref name="directory" /> into <paramref name="archiveFile" />, choosing the compression format based on the
+    ///     archive file's extension.
+    /// </summary>
+    /// <param name="directory">The directory whose contents should be compressed.</param>
+    /// <param name="archiveFile">
+    ///     The archive file to create. Its extension determines the compression format (e.g. <c>.zip</c>,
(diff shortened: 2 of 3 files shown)


ITaluone:
@dennisdoomen This should look familiar to you (did the same thing in FA.. this whole nodejs thing ;) )

ChrisonSimtian:
@ITaluone i wanna bring in some more discipline, can you create an issue for that please and describe the use case we're trying to solve here?
I love the enthusiasm of adding new features but I think we should go the extra mile and help people understand why they want this feature :-)
Probably also needs some docs (which can be a dedicated PR linking to the same issue ;-) see how this becomes a shoe? :D )

And then I'd say lets target this for the next release, not the current one. I wanna close the lid on this one (already relabelled)

Directory.Packages.props
```diff
@@ -26,6 +26,7 @@
     <PackageVersion Include="Serilog.Sinks.Console" Version="6.1.1" />
     <PackageVersion Include="Serilog.Sinks.File" Version="7.0.0" />
     <PackageVersion Include="SharpZipLib" Version="1.4.2" />
+    <PackageVersion Include="SharpCompress" Version="0.50.1" />
```
ChrisonSimtian, line 29: do we want to replace `SharpZipLib` with `SharpCompress`?
I know this is not the scope of this PR, but maybe this could become a consideration of the issue to move from one lib to the other because it give us wider archive support, while not inflating the external lib surface (i.e. replace one with the other instead of introducing an additional one)

ITaluone:
> @ITaluone i wanna bring in some more discipline, can you create an issue for that please and describe the use case we're trying to solve here? I love the enthusiasm of adding new features but I think we should go the extra mile and help people understand why they want this feature :-) Probably also needs some docs (which can be a dedicated PR linking to the same issue ;-) see how this becomes a shoe? :D )
> 
> And then I'd say lets target this for the next release, not the current one. I wanna close the lid on this one (already relabelled)

@ChrisonSimtian 
I don't get it? there is an issue already? Or isn't the issue detailed enough?

Directory.Packages.props
ITaluone: In future maybe.. I have to check whether sharpcompress is capable of everything SharpZipLib is...

ITaluone: @ChrisonSimtian 
Should we add other common archive formats then as well? Like:
- `.gz` (w/o tar)
- `.7z`
- `.tar` (alone)
- (`.rar`) => only decompress
- ...

src/Fallout.Utilities.IO.Compression/CompressionExtensions.cs
```diff
@@ -7,6 +7,8 @@
 using ICSharpCode.SharpZipLib.GZip;
 using ICSharpCode.SharpZipLib.Tar;
 using ICSharpCode.SharpZipLib.Zip;
+using SharpCompress.Common;
+using SharpCompress.Readers;
 using Fallout.Common.Utilities.Collections;
 
 namespace Fallout.Common.IO;
```
dennisdoomen, line 15: 🔧 (required) Please also do an auto-cleanup with Rider or R# so we improve the layout of existing files gradually.

src/Fallout.Utilities.IO.Compression/CompressionExtensions.cs
```diff
@@ -132,6 +138,26 @@ public static void UnTarBZip2To(this AbsolutePath archiveFile, AbsolutePath dire
         UncompressTar(archiveFile, directory, x => new BZip2InputStream(x));
     }
 
+    public static void UnTarXzTo(this AbsolutePath archive, AbsolutePath directory)
```
dennisdoomen, line 246: 🔧 We need to be more rigid on the documentation side, something Nuke really lacked. I'm expecting some XML docs on what does method does, which extensions it supports, what happens if `directory` doesn't exist, what happens if `directory` contains conflicting files, etc.

src/Fallout.Utilities.IO.Compression/Fallout.Utilities.IO.Compression.csproj
```diff
@@ -10,6 +10,7 @@
 
   <ItemGroup>
     <PackageReference Include="SharpZipLib" />
+    <PackageReference Include="SharpCompress" />
```
dennisdoomen, line 13: @ChrisonSimtian do you see a concern with this extra dependency? I don't.

dennisdoomen (changes requested):

src/Fallout.Utilities.IO.Compression/CompressionExtensions.cs
IT-VBFK: I wouldn't hang this up on the extension. When I have e.g. a `.tar.gz` I most definitely could give it an extension: `.blablubb` and would be a valid `.tar.gz`.

SharpCompress.. (the replacement from #610) has a method `ArchiveFactory.IsArchive(string path, out ArchiveType type)` and `ArchiveFactory.GetArchiveInformation(string path);` which I will use there.

What it does? Well.. like the method says: Un-Tars an .tar.xz file to...`directory`

When `directory` not exists.. well this one is valid IMHO to mention

src/Fallout.Utilities.IO.Compression/Fallout.Utilities.IO.Compression.csproj
ChrisonSimtian: like mentioned somewhere else, I'd replace one with the other. Lets stick to one lib. They both overlap :-)

src/Fallout.Utilities.IO.Compression/CompressionExtensions.cs
ChrisonSimtian: +1

Directory.Packages.props
ChrisonSimtian: SharpCompress brings that in for free doesnt it? Then yes

src/Fallout.Utilities.IO.Compression/CompressionExtensions.cs
dennisdoomen: It doesn't help to be sarcastic here. As I said before, one of my goals is to address the lack of proper documentation on a lot of APIs that Nuke suffered from. So please make sure that new methods/APIs make an effort to document the details.

src/Fallout.Utilities.IO.Compression/Fallout.Utilities.IO.Compression.csproj
ITaluone: THis will be a follow up: #610

dennisdoomen (approved):

ITaluone:
Hello @ChrisonSimtian 

Can you do the final review here?

Thanks

ChrisonSimtian:
> Hello @ChrisonSimtian 
> 
> 
> 
> Can you do the final review here?
> 
> 
> 
> Thanks

i'll try but it's a busy week rn and we're changing branching in this repo, so no rush 😅

PR #598, opened 2026-07-29
Add an additional type converter for `System.Text.Json`

### LINKED ISSUE #588: `AddProperty/SetProperty` serialization may produce unexpected results
### Usage Information

Fallout version: 10.4.0-rc.4, SDK version: 10.0.301, TF: net10.0, OS: Windows Server 2019 GitHub Actions runner

### Description

Ran into this migrating a build from Nuke to Fallout: passing an `AbsolutePath` into `AddProperty/SetProperty` on any `ToolOptions`-derived settings (e.g. `DotNetMSBuildSettings`) doesn't give you the path back out — you get a full JSON dump of the `AbsolutePath` object instead. Whatever reads that property afterwards (in my case, the actual MSBuild invocation) chokes on it.

Dug through the source a bit: `AddProperty/SetProperty` end up in `Fallout.Common.Tooling.Options.Set(string, object)`, which does `JsonSerializer.SerializeToNode(value, value.GetType(), SerializerOptions)`. `SerializerOptions (Fallout.Tooling/Options.cs)` registers a handful of converters (`TypeConverter` for `Options` subclasses, `LookupTableConverter`, `EnumerationJsonConverterFactory`), but nothing for `AbsolutePath`.

`AbsolutePath` itself only knows how to turn into a string via a plain old `System.ComponentModel.TypeConverter ([TypeConverter(typeof(TypeConverter))]` on `Fallout.Utilities/IO/AbsolutePath.cs:17`). `System.Text.Json` just doesn't look at those — `Newtonsoft.Json` does (and that's what Nuke's older `Nuke.Tooling.Options` used under the hood), which is presumably why nobody ever hit this on the Nuke side. Without a matching `JsonConverter`, STJ falls back to reflecting over `AbsolutePath`'s public properties (Name, NameWithoutExtension, Extension, Parent, recursively walking all the way up to the drive root) and that's the "dump" you get back instead of the path.

### Reproduction Steps

1. In a Fallout build target, pass an AbsolutePath as a property value:
```csharp
DotNetMSBuild(c => c
    .SetTargetPath(SomeProjectPath)
    .SetTargets("Publish")
    .AddProperty("PublishDir", SomeAbsolutePathDirectory)); // AbsolutePath, not string
```
2. Run the target and inspect the rendered command line.

### Expected Behavior

=> `/property:PublishDir=C:\repo\src\Api\bin\publish\Release`

### Actual Behavior

Actual behavior
`/property:PublishDir="{
  \"Name\": \"Release\",
  \"NameWithoutExtension\": \"Release\",
  \"Extension\": \"\",
  \"Parent\": { ... }
}"`
-> downstream tool (MSBuild) fails: MSB1006: Property is not valid.

### Regression?

Yes. Works correctly under upstream Nuke - Nuke.Tooling/Options.cs uses Newtonsoft.Json (JValue.FromObject), which honors a type's [TypeConverter] and serializes via its string conversion. Fallout's move to System.Text.Json for the same storage dropped that fallback without a replacement converter for AbsolutePath.

### Known Workarounds

Convert explicitly to string before passing it in:
`.AddProperty("PublishDir", SomeAbsolutePathDirectory.ToString())`

Possible upstream fix: register a `JsonConverter` for `AbsolutePath` in `Options.SerializerOptions`, or have the STJ pipeline fall back to a type's TypeConverter the way `Newtonsoft.Json` does.

No

ITaluone:
Add a `System.Text.Json`-aware converter for `AbsolutePath` so `AddProperty`/`SetProperty` serialize it as a plain path string again.

### What changed
- Add a `JsonConverter<AbsolutePath>` (`AbsolutePath.STJTypeConverter`, `#if NET6_0_OR_GREATER`) that reads/writes the path as a JSON string.
- Keep the old `System.ComponentModel.TypeConverter` for pre-.NET 6 targets, since `System.Text.Json` never looks at it — that's the root cause.
- Minor whitespace cleanup in `Equals`.

### Why
`Options.Set` serializes property values with `System.Text.Json`, which ignores `[TypeConverter]`. Without a matching `JsonConverter`, it fell back to reflecting over `AbsolutePath`'s public properties (`Name`, `Parent`, ...), producing a JSON dump instead of the path — breaking anything downstream that expects a plain string (e.g. MSBuild's `/property:`).

Closes #588

### DIFF
--- a/src/Fallout.Utilities/IO/AbsolutePath.cs
+++ b/src/Fallout.Utilities/IO/AbsolutePath.cs
@@ -2,11 +2,14 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
-using System.Globalization;
 using System.IO;
-using System.Linq;
+#if NET6_0_OR_GREATER
+using System.Text.Json.Serialization;
+using Fallout.Utilities.Converters;
+#endif
 using Fallout.Common.Utilities;
 using static Fallout.Common.IO.PathConstruction;
+using TypeConverter = Fallout.Utilities.Converters.TypeConverter;
 
 namespace Fallout.Common.IO;
 
@@ -14,6 +17,9 @@ namespace Fallout.Common.IO;
 /// Represents an absolute path without distinction between files and directories.
 /// </summary>
 [Serializable]
+#if NET6_0_OR_GREATER
+[JsonConverter(typeof(AbsolutePathJsonConverter))]
+#endif
 [TypeConverter(typeof(TypeConverter))]
 [DebuggerDisplay("{" + nameof(path) + "}")]
 public class AbsolutePath : IAbsolutePathHolder, IFormattable
@@ -23,29 +29,6 @@ public class AbsolutePath : IAbsolutePathHolder, IFormattable
     public const string SingleQuote = "s";
     public const string SingleQuoteIfNeeded = "sn";
     public const string NoQuotes = "nq";
-
-    public class TypeConverter : System.ComponentModel.TypeConverter
-    {
-        public override bool CanConvertFrom(ITypeDescriptorContext context, Type sourceType)
-        {
-            return sourceType == typeof(string) || base.CanConvertFrom(context, sourceType);
-        }
-
-        public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
-        {
-            if (value is string stringValue)
-            {
-                return HasPathRoot(stringValue)
-                    ? (AbsolutePath) stringValue
-                    : EnvironmentInfo.WorkingDirectory / stringValue;
-            }
-
-            if (value is null)
-                return null;
-
-            return base.ConvertFrom(context, culture, value);
-        }
-    }
 
     public static AbsolutePath Create(string path)
     {
@@ -149,11 +132,14 @@ public class AbsolutePath : IAbsolutePathHolder, IFormattable
     {
         if (ReferenceEquals(objA: null, obj))
             return false;
+
         if (ReferenceEquals(this, obj))
             return true;
+
         if (obj.GetType() != GetType())
             return false;
-        return Equals((AbsolutePath) obj);
+
+        return Equals((AbsolutePath)obj);
     }
 
     public override int GetHashCode()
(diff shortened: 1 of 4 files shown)


src/Fallout.Utilities/IO/AbsolutePath.cs
```diff
@@ -24,6 +30,33 @@ public class AbsolutePath : IAbsolutePathHolder, IFormattable
     public const string SingleQuoteIfNeeded = "sn";
     public const string NoQuotes = "nq";
 
+#if NET6_0_OR_GREATER
+    public sealed class STJTypeConverter : JsonConverter<AbsolutePath>
```
ChrisonSimtian: lets put this into its own class file please?

ChrisonSimtian:
```suggestion
    public sealed class AbsolutePathJsonConverter : JsonConverter<AbsolutePath>
```

src/Fallout.Utilities/IO/AbsolutePath.cs
```diff
@@ -14,6 +17,9 @@ namespace Fallout.Common.IO;
 /// Represents an absolute path without distinction between files and directories.
 /// </summary>
 [Serializable]
+#if NET6_0_OR_GREATER
+[JsonConverter(typeof(STJTypeConverter))]
```
ChrisonSimtian:
```suggestion
[JsonConverter(typeof(AbsolutePathJsonConverter))]
```

ChrisonSimtian (approved):
lgtm, see code change _**suggestions**_ :-)

ChrisonSimtian:
and if that needs to go into v10.4, which I assume it does(?) then you'd might wanna merge this into the v10.4 branch

ITaluone:
Add to main an cherry-pick is the better way, IMHO

Backporting is done like this every time in my day work

dennisdoomen:
@ITaluone why did you request a review and then merged it after all?

IT-VBFK:
Hmm.. then we can scratch the "Can merge" label? Otherwise, I cannot tell the value of it TBH.

I added both of you, because "the quicker wins", but if that's not how it works, I will add just one next time.

PR #610, opened 2026-07-30
Replace `SharpZipLib` with `SharpCompress`

ITaluone:
Replaces the library `SharpZipLib` with `SharpCompress`.

### Why?
- `SharpCompress` has more features compared to `SharpZipLib` (especially the xz decompression)
- No need to maintain two compressing libraries

### Verification
Added the tests **before** doing anything and saving the directory hash from the "old" state, then replaced the compression lib and all tests stay green.

### Request
Per https://github.com/Fallout-build/Fallout/pull/597#discussion_r3679159474

### DIFF
--- a/src/Fallout.Utilities.IO.Compression/CompressionExtensions.cs
+++ b/src/Fallout.Utilities.IO.Compression/CompressionExtensions.cs
@@ -3,14 +3,11 @@ using System.Collections.Generic;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
-using Fallout.Common.Utilities.Collections;
-using ICSharpCode.SharpZipLib.BZip2;
-using ICSharpCode.SharpZipLib.GZip;
-using ICSharpCode.SharpZipLib.Tar;
-using ICSharpCode.SharpZipLib.Zip;
 using SharpCompress.Common;
 using SharpCompress.Readers;
-using ZipFile = ICSharpCode.SharpZipLib.Zip.ZipFile;
+using SharpCompress.Writers;
+using SharpCompress.Writers.Zip;
+using Fallout.Common.Utilities.Collections;
 
 namespace Fallout.Common.IO;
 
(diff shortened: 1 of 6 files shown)


build/Build.Compression.cs
```diff
@@ -0,0 +1,140 @@
+using System;
+using System.IO;
+using System.IO.Compression;
+using System.Text;
+using Fallout.Common;
+using Fallout.Common.IO;
+
+partial class Build
```
dennisdoomen: 🤔 Why do we need all of this? Can't you cover the new APIs with simple unit tests?

IT-VBFK: Hmm.. true.. I first thought: "how could I verify, that the files are the same before and after compression?", but now I found the `AbsolutePath.GetDirectoryHash()` extension.

This can work I think:

```csharp

using System;
using System.Text;
using Fallout.Common.IO;
using FluentAssertions;
using Xunit;

namespace Fallout.Common.Specs;

public class CompressionSpecs : IDisposable
{
    private readonly AbsolutePath tempDirectory = AbsolutePath.Temp(nameof(CompressionSpecs));
    private readonly AbsolutePath tempDirectory2 = AbsolutePath.Temp(nameof(CompressionSpecs));

    public CompressionSpecs()
    {
        GenerateFiles(tempDirectory, 100, 15);
    }

    [Fact]
    public void A_directory_hash_is_the_same_before_and_after_compression()
    {
        var hash = tempDirectory.GetDirectoryHash();
        tempDirectory.TarBZip2To(tempDirectory / "comp.tar.bz2");

        (tempDirectory / "comp.tar.bz2").UnTarBZip2To(tempDirectory2);

        tempDirectory2.GetDirectoryHash().Should().Be(hash);
    }

    public void Dispose()
    {
        tempDirectory.DeleteDirectory();
        tempDirectory2.DeleteDirectory();
    }

    private static void GenerateFiles(AbsolutePath rootDirectory, int fileCount, int maximumDirectoryDepth) { ... }
}
```

I'll change this after vacation :)

build/Build.cs
ChrisonSimtian, line 1: nit: the reformat is a bit irritating for the reviewers, I'd try to avoid that in the future. We can just make a 2nd PR for that, especially since the actual change here is just 2 additional lines of code

src/Fallout.Utilities.IO.Compression/CompressionExtensions.cs
```diff
@@ -50,36 +50,23 @@ public static void ZipTo(
...
     {
-        using var fileStream = File.OpenRead(archiveFile);
-        using var zipFile = new ICSharpCode.SharpZipLib.Zip.ZipFile(fileStream);
-
-        var entries = zipFile.Cast<ZipEntry>().Where(x => !x.IsDirectory);
-
-        void HandleEntry(ZipEntry entry)
-        {
-            var file = directory / entry.Name;
-            Directory.CreateDirectory(file.Parent.NotNull());
```
ChrisonSimtian, line 142: @IT-VBFK this line here creates empty directories. the new code doesnt seem to handle this. Can we make sure the behaviour stays the same or at least document why it doesnt matter?
Pretty sure Matt did this on purpose when he wrote this :-)

ChrisonSimtian (commented):
lmk when you need a final review ;-) Enjoy your summer vacation :-)

src/Fallout.Utilities.IO.Compression/CompressionExtensions.cs
ITaluone: I am testing this right now. This is actually a no-op, because empty directories are not even created in zip files.

Verified manually via:

```csharp
[Fact]
public void Test()
{
    var path = AbsolutePath.Temp() / "a" / "b" / "c";
    path.CreateDirectory();
    var archivePath = AbsolutePath.Temp() / "archive.zip";
    path.CompressTo(archivePath);
}
```

and then opend the zip file => fully empty.

ITaluone:
@ChrisonSimtian After a long time, here we go again. This should be it now.

ChrisonSimtian (approved):

dennisdoomen (approved):

PR #661, opened 2026-08-27
Add an llms.txt index for the documentation (#660)

### LINKED ISSUE #660: Add llms.txt for LLM support
## Description

Publish an [`llms.txt`](https://llmstxt.org/) index at `https://docs.fallout.build/llms.txt` so an LLM can find Fallout's documentation without crawling and guessing.

The 37 pages under `docs/website/` are the source of truth for `docs.fallout.build`, which is built by the separate [`Fallout-build/docs.fallout.build`](https://github.com/Fallout-build/docs.fallout.build) Docusaurus repo. Today a model answering a question about Fallout has three bad options: crawl the rendered HTML and spend most of its tokens on navigation chrome, fall back on what it learned about NUKE and get the namespaces and the tool name wrong, or guess. The rebrand makes the third option actively harmful, because a model's training data almost certainly holds `Nuke.*` and `nuke :setup` rather than `Fallout.*` and `fallout :setup`.

`llms.txt` is a small markdown file at the site root: a title, a one-line summary, and grouped lists of links with descriptions. It costs one fetch and gives a model an accurate map of what exists and where.

The file must not be hand-maintained. This repo already has the right pattern for a generated artifact: `GenerateTools` writes the `.Generated.cs` wrappers and `VerifyGeneratedTools` fails CI when a spec was edited without regenerating. `llms.txt` should work the same way, so a page added to `docs/website/` cannot silently go missing from the index.

## Usage Example

An AI assistant answering "how do I install Fallout" fetches one file:

```
GET https://docs.fallout.build/llms.txt
```

```markdown
# Fallout

> A C#-first build automation framework for .NET, the hard-fork successor to NUKE.
> Write CI/CD pipelines in plain C#, debug them locally, and share build steps
> across repositories.

## Getting Started

- [Installation](https://docs.fallout.build/getting-started/installation): Install the Fallout global tool.
- [Build Setup](https://docs.fallout.build/getting-started/setup): Set up a build project in a repository.
- [Build Execution](https://docs.fallout.build/getting-started/execution): Run and debug targets locally.

## Fundamentals

- [Targets](https://docs.fallout.build/fundamentals/targets): Declare targets and dependencies.
...
```

It then fetches only the one page it needs, and answers with `dotnet tool install Fallout.GlobalTool --global` instead of the NUKE command it was trained on.

A contributor regenerates the file the same way they regenerate a tool wrapper:

```powershell
./build.ps1 GenerateLlmsTxt
```

## Alternative

Today there is none. A model has to crawl `docs.fallout.build` page by page, or answer from stale NUKE knowledge. Neither is reliable, and the second one is wrong in exactly the places the rebrand changed.

Hand-writing the file in the Docusaurus repo is the obvious shortcut, and it is what this issue argues against. See the Brainstorm below.

## Could you help with a pull-request?

Yes

**Related:** #648 (AI-native build interface epic), #646 (MCP server), #240 (project-local AI tooling config), #104 (AI-friendly structured output modes)

These cover making the **build engine** machine-readable. This issue covers making the **documentation** machine-readable. They are neighbours, not the same job: nothing in #648's scope table would produce an `llms.txt`, and resolving #648 would not resolve this.

<details>
<summary><b>🧠 Brainstorm</b></summary>

### Problem and context

- `docs/website/` holds 37 markdown pages. Every one carries a `title:` in its frontmatter; some also carry a `description:`.
- Directory and file names are numbered for ordering (`01-getting-started/01-installation.md`). Docusaurus strips the `NN-` prefixes, so that page is served at `https://docs.fallout.build/getting-started/installation`. The README already links that exact URL, which confirms the mapping.
- The markdown lives here. The site that serves it lives in `Fallout-build/docs.fallout.build`. Whatever this issue produces has to cross that boundary.
- `AGENTS.md` says the build is itself a C# console app and is the canonical example of consuming the framework. Generating this file with a Fallout target dogfoods the framework rather than adding a second toolchain.

### Approaches

**A. Hand-write a static `llms.txt` in the Docusaurus repo.**

Cheapest to start: one file, no build changes, done in an afternoon.

It rots on the first docs PR. The person adding a page to `docs/website/` in this repo has no reason to open the site repo, so the index silently falls behind the docs it indexes. A stale map is worse than no map, because a model trusts it and then fetches a URL that moved. This is also the one option that contradicts the repo's own convention for generated artifacts.

**B. Generate from `docs/website/` frontmatter with a Fallout target, commit the result, gate it in CI.**

Mirrors `GenerateTools` / `VerifyGeneratedTools` exactly, which is the pattern this repo already uses for exactly this problem. The generator reads the frontmatter that is already there, so it needs no new metadata and no new dependency. The artifact is committed, so it shows up in a diff and a reviewer can read what changed. `VerifyLlmsTxt` fails the build when someone edits a page without regenerating, which is the property option A cannot have.

Costs: a checked-in generated file, and one target more in the build. The site repo still needs a one-line change to serve the file at the root.

**C. A Docusaurus plugin in the site repo that emits `llms.txt` at site build time.**

Always matches what is actually published, and nothing is checked in.

The logic ends up in the repo that does not own the content, written in a different language, invisible to the people editing the docs. It cannot be reviewed in the same diff as the page that changed, and it makes the site repo, currently a thin renderer, into a place with its own behaviour to maintain. It also fails the `AGENTS.md` preference for dogfooding.

### Recommendation

**Option B.**

The deciding argument is the drift, not the effort. All three options produce a correct file on day one, and only B is still correct on day one hundred. B is also the option that already has a working precedent in this repo, so it introduces a target and a gate that a reviewer here recognises on sight rather than a new mechanism they have to evaluate.

The known cost of B is the cross-repo hop: the generated file has to reach the site root. That is one static-asset copy in the Docusaurus repo and it is called out as a non-goal below, with a companion issue to file.

</details>

<details>
<summary><b>📋 Spec</b></summary>

### Goal

`https://docs.fallout.build/llms.txt` serves a valid [llmstxt.org](https://llmstxt.org/) index of every published page, regenerated by `./build.ps1 GenerateLlmsTxt` and guarded by a CI gate that fails when the file is out of sync with `docs/website/`.

### Scope

- A `GenerateLlmsTxt` target that walks `docs/website/**/*.md` and writes `docs/llms.txt`.
- A `VerifyLlmsTxt` target that fails when the committed file is stale, in the shape of `VerifyGeneratedTools`.
- The generated `docs/llms.txt`, committed.
- Wiring `VerifyLlmsTxt` into the CI build so the gate actually runs.

### Non-goals

- **`llms-full.txt`** (the whole corpus inlined). Worth having, and cheap once the walk exists, but it is a different file with a different size budget. Separate issue.
- **Serving the file.** `static/llms.txt` in `Fallout-build/docs.fallout.build` is a one-line change in a repo this issue does not touch. File a companion issue there and link it; this issue is not done in the user-visible sense until that lands, and the PR description should say so.
- **Indexing anything outside `docs/website/`.** `docs/adr/`, `docs/agents/` and `AGENTS.md` are contributor-facing and are not published on the site. Linking a URL that 404s is worse than omitting it.
- **A new metadata field in the frontmatter.** The generator uses what is there.

### Pipeline

```mermaid
flowchart LR
    A["docs/website/**/*.md<br/>37 pages"] --> B["Parse frontmatter<br/>title, description"]
    B --> C["Derive URL<br/>strip NN- prefixes"]
    C --> D["Group by section<br/>from directory"]
    D --> E["Render llms.txt"]
    E --> F["docs/llms.txt<br/>committed"]
    F -. "companion issue" .-> G["docs.fallout.build<br/>static/llms.txt"]
    F --> H["VerifyLlmsTxt<br/>CI gate"]
```

### URL derivation

Given a path relative to `docs/website/`, strip the `NN-` prefix from every segment, drop the `.md` extension, and prefix `https://docs.fallout.build/`.

| Source path | URL |
|---|---|
| `01-getting-started/01-installation.md` | `.../getting-started/installation` |
| `05-cicd/03-github-actions.md` | `.../cicd/github-actions` |
| `badge.md` | `.../badge` |

### Section grouping and ordering

Sections come from the top-level directory, title-cased from the slug with the numeric prefix stripped: `01-getting-started` becomes `Getting Started`. Sections and the pages inside them keep the numeric order the prefixes encode, so the file reads in the same order as the site sidebar. Root-level pages (`badge.md`) go in a trailing `## Optional` section, which is the part of the llmstxt.org format meant for lower-priority links.

### Page entries

`- [{title}]({url}): {description}` where `title` is the frontmatter `title`. When `description` is absent, use the page's first non-empty prose line, trimmed to roughly 200 characters. Skip Docusaurus `import` lines and MDX components when picking that line, since several pages open with one.

### Header

The H1 is `# Fallout`. The blockquote summary reuses `docs/website/introduction.md`'s frontmatter `description`, which already reads as a one-line pitch, so the summary has exactly one source and cannot disagree with the site.

### Edge cases

- **`introduction.md` has `sidebar_position: 0`** and may be served at `/` rather than `/introduction`. Confirm against the live site before generating, and special-case it if so.
- **A page with no `title:`** should fail the target loudly rather than emit a blank link. That is a docs bug worth surfacing.
- **Line endings.** The repo is on Windows with CRLF in the working tree. Write the file with the same convention the other generated artifacts use, or `VerifyLlmsTxt` will report drift on every clean checkout.

### Assumptions

- `docs.fallout.build` serves `static/` at the site root, which is the Docusaurus default.
- 37 pages produce a file in the low tens of kilobytes, well inside any practical context budget, so no truncation is needed.
- **There is no test project for `build/`, so the verification model is the `Generate` plus `Verify` target pair this repo already uses for `.Generated.cs`, not a unit-test suite.** The plan below is shaped accordingly: each task ends by running the target and reading the output, and the CI gate is the durable regression guard. Adding a `tests/Build.Specs` project to unit-test the generator would be a defensible alternative, but it is a bigger change than this issue warrants and it should be its own decision.

</details>

## 🛠️ Implementation plan

> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.

**Goal.** Generate and CI-gate an llmstxt.org index of `docs/website/`, following the `GenerateTools` / `VerifyGeneratedTools` pattern.

**Architecture.** The build is a C# console app in `build/`. New targets go in a partial of `Build` in their own `build/Build.*.cs` file, matching `Build.CodeGeneration.cs`.

**Tech stack.** .NET 10 (`global.json`, `rollForward: latestMinor`), Fallout's own `AbsolutePath` / `GlobFiles` / `Assert` helpers.

**Global constraints.**

- **No conventional commits.** `AGENTS.md` rule 7 forbids `feat:`, `fix:`, `chore:` and the rest. Commit messages are functional descriptions, for example `Generate an llms.txt index from the docs frontmatter`.
- **No per-file license headers** (`AGENTS.md` rule 6).
- **Terse plain English** in commits and comments (`AGENTS.md` rule 8). One idea per sentence, no idioms.
- **Central package versions only.** If a dependency is needed, it goes in `Directory.Packages.props` with no inline `Version=`. Prefer no new dependency: the frontmatter here is simple enough to parse without a YAML library, and `Fallout.Utilities.Text.Yaml` already exists if it is not.
- **Non-breaking, targets `develop`.** Additive only, no public API change.
- Verify with `./build.ps1 Compile` before each commit.

### Task 1: Walk the docs and derive URLs

**Files:** create `build/Build.Documentation.cs`; no other file changes yet.

**Interfaces:** `Target GenerateLlmsTxt`; private helpers `AbsolutePath DocsWebsiteDirectory`, `string ToPublicUrl(AbsolutePath page)`, `string ToSectionTitle(string directorySlug)`.

- [x] **Step 1:** Create `build/Build.Documentation.cs` as a partial of `Build`, with `DocsWebsiteDirectory => RootDirectory / "docs" / "website"`.
- [x] **Step 2:** Implement `ToPublicUrl`: strip the `NN-` prefix from each path segment, drop `.md`, prefix `https://docs.fallout.build/`.
- [x] **Step 3:** Implement `ToSectionTitle`: strip the numeric prefix and title-case the slug, so `01-getting-started` becomes `Getting Started`.
- [x] **Step 4:** Add a temporary `GenerateLlmsTxt` target that globs the pages and logs each source path next to its derived URL.
- [x] **Step 5:** Run `./build.ps1 GenerateLlmsTxt`. Confirm 37 pages are found, and that `01-getting-started/01-installation.md` maps to `https://docs.fallout.build/getting-started/installation`, which is the URL the README already links.
- [x] **Step 6:** Spot-check three of the logged URLs against the live site, including one from `05-cicd/` and the root-level `badge.md`.
- [x] **Step 7:** Commit: `Walk the docs website and derive public URLs`.

### Task 2: Parse frontmatter into title and description

**Files:** modify `build/Build.Documentation.cs`.

**Interfaces:** `sealed record DocPage(string Title, string Description, string Url, string Section, int Order)`; `DocPage ReadPage(AbsolutePath file)`.

- [x] **Step 1:** Implement `ReadPage`: read the leading `---` frontmatter block and pull `title` and `description`.
- [x] **Step 2:** Fall back to the first non-empty prose line when `description` is missing. Skip `import` lines and MDX component tags, since several pages open with one. Trim to roughly 200 characters on a word boundary.
- [x] **Step 3:** Make a missing `title` fail the target with `Assert`, naming the offending file. A blank link is a docs bug worth surfacing.
- [x] **Step 4:** Extend the target to log the parsed `DocPage` for every file.
- [x] **Step 5:** Run `./build.ps1 GenerateLlmsTxt`. Confirm every page has a non-empty title, and check the fallback description on a page that has no `description:`, for example `01-getting-started/01-installation.md`.
- [x] **Step 6:** Confirm the target fails with a useful message when a `title:` is temporarily removed from a page, then restore it.
- [x] **Step 7:** Commit: `Parse the docs frontmatter into page metadata`.

### Task 3: Render and write docs/llms.txt

**Files:** modify `build/Build.Documentation.cs`; create `docs/llms.txt`.

**Interfaces:** `string RenderLlmsTxt(IReadOnlyList<DocPage> pages)`.

- [x] **Step 1:** Render the header: `# Fallout`, then a blockquote from `docs/website/introduction.md`'s frontmatter `description`.
- [x] **Step 2:** Group pages by section, order sections and pages by their numeric prefixes, and emit `## {Section}` followed by `- [{title}]({url}): {description}` lines.
- [x] **Step 3:** Emit root-level pages such as `badge.md` under a trailing `## Optional` section.
- [x] **Step 4:** Write the result to `docs/llms.txt`, matching the line-ending convention the other generated artifacts use so the gate in Task 4 does not report drift on a clean checkout.
- [x] **Step 5:** Run `./build.ps1 GenerateLlmsTxt`. Read `docs/llms.txt` in full and confirm it is a valid llmstxt.org document: one H1, one blockquote, `##` sections, one link line per page, 37 links.
- [x] **Step 6:** Confirm `introduction.md` resolves to the URL the live site actually serves. If it is served at `/` rather than `/introduction`, special-case it and note the reason in a comment.
- [x] **Step 7:** Run `./build.ps1 GenerateLlmsTxt` a second time and confirm `git status` is clean, which proves the output is deterministic. A non-deterministic generator would make the Task 4 gate flap.
- [x] **Step 8:** Commit: `Generate an llms.txt index from the docs frontmatter`, including the generated `docs/llms.txt`.

### Task 4: Gate the generated file in CI

**Files:** modify `build/Build.Documentation.cs`; modify the CI target list wherever `VerifyGeneratedTools` is wired in.

**Interfaces:** `Target VerifyLlmsTxt`, with `.Requires(() => GitHasCleanWorkingCopy()).DependsOn(GenerateLlmsTxt)`.

- [x] **Step 1:** Add `VerifyLlmsTxt` modelled on `VerifyGeneratedTools` in `build/Build.CodeGeneration.cs`, including the same explanatory comment about why `Requires` fires before the plan runs.
- [x] **Step 2:** Make the failure message point at the fix: `docs/llms.txt is out of sync with docs/website. Run './build.ps1 GenerateLlmsTxt' locally and commit the result.`
- [x] **Step 3:** Find where `VerifyGeneratedTools` is invoked in CI and wire `VerifyLlmsTxt` in the same place, so the gate actually runs on a PR.
- [x] **Step 4:** Run `./build.ps1 VerifyLlmsTxt` on a clean tree and confirm it passes.
- [x] **Step 5:** Edit a page's `title:`, re-run `./build.ps1 VerifyLlmsTxt`, and confirm it fails with the message from Step 2. Restore the page.
- [x] **Step 6:** Run `./build.ps1 Compile` and confirm exit 0 with no new warnings.
- [x] **Step 7:** Commit: `Fail the build when llms.txt is out of sync with the docs`.

### Task 5: Hand off the serving half

**Files:** none in this repo.

- [x] **Step 1:** Confirm `Fallout-build/docs.fallout.build` serves `static/` at the site root, which is the Docusaurus default.
- [x] **Step 2:** File a companion issue on `Fallout-build/docs.fallout.build`: copy `docs/llms.txt` from this repo into `static/llms.txt` during the site build, so the file is served at `https://docs.fallout.build/llms.txt`.
- [x] **Step 3:** Link that issue here, and say plainly in the PR description that this repo's half generates the file while the site repo's half serves it. Neither half is user-visible alone.
- [ ] **Step 4:** Once both have landed, fetch `https://docs.fallout.build/llms.txt` and confirm it serves as `text/plain` or `text/markdown` and matches the committed file.

phmatray:
Implements #660.

Closes #660.

Generates `docs/llms.txt`, an [llmstxt.org](https://llmstxt.org/) index of the 37 pages under `docs/website/`, and fails the build when it goes stale.

`llms.txt` lets an LLM answer questions about Fallout from one fetch instead of crawling the site or falling back on stale NUKE knowledge. That fallback is the part that matters after the rebrand: a model trained before it will say `Nuke.*` and `nuke :setup`.

### What is here

| File | Change |
|---|---|
| `build/Build.Documentation.cs` *(new)* | `GenerateLlmsTxt` and `VerifyLlmsTxt` |
| `docs/llms.txt` *(new, generated)* | the index itself, committed |
| `build/Build.CI.GitHubActions.cs` | `VerifyLlmsTxt` added to the `build` workflow's targets |
| `.github/workflows/build.yml` | regenerated from the attribute above |
| `.github/workflows/build-skip.yml` | runs `VerifyLlmsTxt` instead of echoing |
| `.fallout/build.schema.json` | regenerated for the two new targets |

Not hand-maintained, by design. It follows the `GenerateTools` / `VerifyGeneratedTools` pattern already in this repo: generate, commit, and fail CI when a source edit lands without regenerating.

### Decisions worth reviewing

**The gate is wired into both workflows, and that is the point.** `build.yml` has `paths-ignore: docs/**`, so a PR editing a page under `docs/website/` never runs it. That is exactly the change that invalidates `llms.txt`. Wiring the gate only there would have produced a check that can never fire on the thing it guards. `build-skip.yml` is the workflow that handles those PRs, so it now runs `VerifyLlmsTxt` rather than an `echo`. It keeps the `ubuntu-latest` job name that branch protection keys on, and it builds the build project and regenerates one file rather than running test and pack.

**Metadata comes from what the site already declares, not from new fields.** Section names are read from each directory's Docusaurus `_category_.json` `label`, so they read "CI/CD Support" and "Common Tasks" exactly as the sidebar does. Title-casing the slug instead would have produced "Cicd" and "Ide". Titles fall back to the first H1 when a page declares no `title`, which is what Docusaurus does and what `badge.md` relies on. Descriptions fall back to the page's opening paragraph, since only `introduction.md` declares one.

**The base URL is `https://docs.fallout.build/docs/`,** verified against the live `sitemap.xml`. All 37 derived URLs match the 37 live pages exactly, set against set. The `/docs/` prefix is easy to miss: see the follow-up below.

### Test plan

- [x] All 37 derived URLs compared against `https://docs.fallout.build/sitemap.xml`. Identical in both directions, so no derived URL 404s and no live page is missing.
- [x] `./build.ps1 GenerateLlmsTxt` twice in a row leaves `git status` clean. The generator is deterministic, which is what keeps the gate from flapping.
- [x] `./build.ps1 VerifyLlmsTxt` on a clean tree exits 0.
- [x] Negative test: edited a page's `title:`, committed without regenerating, and the gate failed with `docs/llms.txt is out of sync with docs/website. Run './build.ps1 GenerateLlmsTxt' locally and commit the result.` Reverted afterwards.
- [x] Negative test: removed a page's `title:` and the target failed naming the offending file.
- [x] `./build.ps1 VerifyGeneratedTools VerifyLlmsTxt Test`, the same target set CI runs: 814 passed, 7 skipped, 0 failed, and the working tree stayed clean, so both verify gates passed.
- [x] `./build.ps1 Compile` exits 0. 8 warnings, all pre-existing `NU190x` advisories, none in the new file.
- [x] Generated file checked by hand: one H1, one blockquote, 8 `##` sections, 37 link lines.

### Review round

A high-effort review raised 12 findings. Ten are fixed in `7383c59e`; none changes the generated file today, and each closes a way it could go wrong on a page the docs do not have yet. The two that mattered most:

- `_category_.json` without a `label` **crashed** the target. `GetPropertyValue` throws when the property is absent, so the guard and the slug fallback under it were dead code.
- Underscore-prefixed files were indexed. Docusaurus excludes them from routing and `docs/website/_snippets/` already exists, so a `.md` dropped there would have emitted a URL the site never serves.

Also fixed: `.mdx` pages are indexed, `sidebar_position` orders pages inside a section as it does on the site, titles are read outside code fences, unclosed frontmatter fails instead of shipping garbage, and titles sort with `StringComparer.Ordinal` so a different culture cannot fail the byte-for-byte gate with no real drift.

Verified empirically rather than by reading: added `_snippets/partial.md`, a `.mdx` page with `sidebar_position: 1`, and a label-less `_category_.json`, regenerated, and confirmed the snippet was skipped, the MDX page indexed and sorted first, and the label-less category fell back to its slug instead of throwing. Reverted afterwards; `docs/llms.txt` is byte-identical.

**One finding dismissed.** The review flagged that `docs/llms.txt` sits outside `docs/website/` and might never be published. That path is the deliberate cross-repo contract, not an oversight: it is an artifact *of* the website sources rather than a page *in* them, and [Fallout-build/docs.fallout.build#14](https://github.com/Fallout-build/docs.fallout.build/issues/14) names the exact path to copy. Worth confirming with whoever owns the site build.

### Serving half is a separate repo

`docs.fallout.build` is built from [`Fallout-build/docs.fallout.build`](https://github.com/Fallout-build/docs.fallout.build). This PR generates the file; that repo has to copy it to `static/llms.txt`. Filed as [Fallout-build/docs.fallout.build#14](https://github.com/Fallout-build/docs.fallout.build/issues/14).

**Neither half is user-visible alone.** `https://docs.fallout.build/llms.txt` stays 404 until both land. That is expected, not an oversight.

### Plan

- [x] Task 1: Walk the docs and derive URLs
- [x] Task 2: Parse frontmatter into title and description
- [x] Task 3: Render and write docs/llms.txt
- [x] Task 4: Gate the generated file in CI
- [x] Task 5: Hand off the serving half

## Follow-ups

Not fixed here, to keep this PR to its issue.

- **`README.md:59` links to `https://docs.fallout.build/getting-started/installation`, which 404s.** It is missing the `/docs/` prefix. Found while verifying the URL derivation: the issue cited this link as proof of the mapping, and it turned out to be the stale one. It is the only such link in the repo.
- **`llms-full.txt`** (the corpus inlined rather than linked). Cheap now that the walk exists, but a different file with a different size budget.
- **`build-skip.yml` now has a real build step**, which makes it the natural home for other docs-only validation, such as link checking. Nothing is broken; it is just no longer true that docs PRs run nothing.

### DIFF
--- a/.github/workflows/build-skip.yml
+++ b/.github/workflows/build-skip.yml
@@ -28,8 +28,13 @@ name: build
 # Mirrors build.yml. This job does real work now, so rapid pushes to a docs PR
 # would otherwise stack concurrent builds, which is the cost this file exists to
 # avoid in the first place.
+#
+# The group is a literal, NOT ${{ github.workflow }}: that expands to the workflow
+# `name:`, which is `build` in both this file and build.yml on purpose. A PR that
+# touches code AND a doc fires both workflows, and a shared group plus
+# cancel-in-progress would let whichever starts second cancel the other.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: build-skip-${{ github.ref }}
   cancel-in-progress: true
 
 on:
--- a/.github/workflows/build-skip.yml
+++ b/.github/workflows/build-skip.yml
@@ -25,6 +25,13 @@
 
 name: build
 
+# Mirrors build.yml. This job does real work now, so rapid pushes to a docs PR
+# would otherwise stack concurrent builds, which is the cost this file exists to
+# avoid in the first place.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 on:
   pull_request:
     branches:
--- a/.github/workflows/build-skip.yml
+++ b/.github/workflows/build-skip.yml
@@ -9,10 +9,15 @@
 #   - Without a substitute, docs-only PRs sit BLOCKED waiting for a check that
 #     never reports.
 #
-# This workflow fires on the inverse path set (docs-only changes), runs nothing
-# of substance, and reports success under the same `ubuntu-latest` status-check
-# context — satisfying the protection rule without spending CI minutes on a real
-# build.
+# This workflow fires on the inverse path set (docs-only changes) and reports
+# success under the same `ubuntu-latest` status-check context, satisfying the
+# protection rule without spending CI minutes on a full build/test/pack.
+#
+# It is not a no-op any more. docs/llms.txt is generated from docs/website by
+# `GenerateLlmsTxt`, so a docs-only PR is exactly the change that can leave it
+# stale — and it is exactly the change build.yml ignores. VerifyLlmsTxt therefore
+# runs here, which is the only workflow that sees these PRs. It builds the build
+# project and regenerates one file; it does not run the test or pack targets.
 #
 # Keep the job name `ubuntu-latest` aligned with build.yml so both files produce
 # a status check named `ubuntu-latest`; the workflow `name:` mirrors build.yml's
@@ -37,5 +42,23 @@ jobs:
     name: ubuntu-latest
     runs-on: ubuntu-latest
     steps:
-      - name: 'Skip: docs-only PR, no build needed'
-        run: echo "Docs-only change — ubuntu-latest validation skipped via .github/workflows/build-skip.yml."
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+          repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
+          ref: ${{ github.head_ref }}
+      - name: 'Cache: .fallout/temp, ~/.nuget/packages'
+        uses: actions/cache@v6
+        with:
+          path: |
+            .fallout/temp
+            ~/.nuget/packages
+          key: ${{ runner.os }}-${{ hashFiles('**/global.json', '**/*.csproj', '**/Directory.Packages.props') }}
+      - name: 'Setup: .NET SDK'
+        uses: actions/setup-dotnet@v6
+        with:
+          global-json-file: global.json
+      - name: 'Restore: dotnet tools'
+        run: dotnet tool restore
+      - name: 'Run: VerifyLlmsTxt'
+        run: dotnet fallout VerifyLlmsTxt
--- a/build/Build.Documentation.cs
+++ b/build/Build.Documentation.cs
@@ -1,0 +1,66 @@
+using System;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Fallout.Common;
+using Fallout.Common.IO;
+using Fallout.Common.Utilities;
+using Fallout.Common.Utilities.Collections;
+using Serilog;
+
+partial class Build
+{
+    AbsolutePath DocsWebsiteDirectory => RootDirectory / "docs" / "website";
+
+    // The public site is built from docs/website by the separate Fallout-build/docs.fallout.build
+    // Docusaurus repository, which serves the pages under a /docs/ route prefix. Verified against
+    // https://docs.fallout.build/sitemap.xml, not against the README: the README's own links omit
+    // the prefix and 404 (see the follow-up on the PR). Hardcoded for the same reason as
+    // CanonicalRepositoryIdentifier: the generated file must be identical whichever fork
+    // regenerates it.
+    const string DocsBaseUrl = "https://docs.fallout.build/docs/";
+
+    // Docusaurus orders pages by a numeric prefix on the directory and file name, and strips that
+    // prefix from the served URL. So 01-getting-started/01-installation.md is served at
+    // /getting-started/installation, which is the URL the README already links.
+    static readonly Regex OrderPrefix = new(@"^(?<order>\d+)-", RegexOptions.Compiled);
+
+    static string StripOrderPrefix(string segment) => OrderPrefix.Replace(segment, string.Empty);
+
+    static int GetOrder(string segment)
+    {
+        var match = OrderPrefix.Match(segment);
+        return match.Success ? int.Parse(match.Groups["order"].Value) : int.MaxValue;
+    }
+
+    string ToPublicUrl(AbsolutePath page)
+    {
+        var relative = DocsWebsiteDirectory.GetUnixRelativePathTo(page).ToString();
+        var slug = relative[..^".md".Length]
+            .Split('/')
+            .Select(StripOrderPrefix)
+            .JoinSlash();
+
+        return DocsBaseUrl + slug;
+    }
+
+    // "01-getting-started" becomes "Getting Started". The directory slug is the only section name
+    // available: the docs tree has no per-directory metadata file.
+    static string ToSectionTitle(string directorySlug)
+    {
+        return StripOrderPrefix(directorySlug)
+            .Split('-')
+            .Select(x => char.ToUpperInvariant(x[0]) + x[1..])
+            .JoinSpace();
+    }
+
+    Target GenerateLlmsTxt => _ => _
+        .Executes(() =>
+        {
+            DocsWebsiteDirectory.GlobFiles("**/*.md")
+                .OrderBy(x => x.ToString())
+                .ForEach(x => Log.Information(
+                    "{Relative} -> {Url}",
+                    DocsWebsiteDirectory.GetUnixRelativePathTo(x),
+                    ToPublicUrl(x)));
+        });
+}
--- a/build/Build.Documentation.cs
+++ b/build/Build.Documentation.cs
@@ -8,12 +8,17 @@ using Fallout.Common;
 using Fallout.Common.IO;
 using Fallout.Common.Utilities;
 using Fallout.Common.Utilities.Collections;
+using Fallout.Utilities.Text.Yaml;
 using Serilog;
+using YamlDotNet.Serialization;
+using YamlDotNet.Serialization.NamingConventions;
 using static Fallout.Common.Tools.Git.GitTasks;
 
 partial class Build
 {
     AbsolutePath DocsWebsiteDirectory => RootDirectory / "docs" / "website";
+
+    AbsolutePath LlmsTxtFile => RootDirectory / "docs" / "llms.txt";
 
     // The public site is built from docs/website by the separate Fallout-build/docs.fallout.build
     // Docusaurus repository, which serves the pages under a /docs/ route prefix. Verified against
@@ -23,17 +28,138 @@ partial class Build
     // regenerates it.
     const string DocsBaseUrl = "https://docs.fallout.build/docs/";
 
+    const int MaxDescriptionLength = 200;
+
     // Docusaurus orders pages by a numeric prefix on the directory and file name, and strips that
     // prefix from the served URL. So 01-getting-started/01-installation.md is served at
     // /docs/getting-started/installation.
-    static readonly Regex OrderPrefix = new(@"^(?<order>\d+)-", RegexOptions.Compiled);
+    static readonly Regex orderPrefix = new(@"^(?<order>\d+)-", RegexOptions.Compiled);
 
-    static string StripOrderPrefix(string segment) => OrderPrefix.Replace(segment, string.Empty);
+    // Inline markdown links render as "[text](url)". Only the text belongs in a one-line summary.
+    static readonly Regex inlineLink = new(@"\[(?<text>[^\]]+)\]\([^)]+\)", RegexOptions.Compiled);
 
-    static int GetOrder(string segment)
+    // Deserialized rather than hand-parsed, so quoting, escaping and block scalars follow the YAML
+    // spec instead of a line regex. Underscored, not the repo-default camelCase builder, because
+    // Docusaurus spells its keys 'sidebar_position'.
+    static readonly DeserializerBuilder frontmatterDeserializer = new DeserializerBuilder()
+        .WithNamingConvention(UnderscoredNamingConvention.Instance)
+        .IgnoreUnmatchedProperties();
+
+    /// <param name="IsPrimary">
+    /// Whether the site places this page in its sidebar. Only meaningful for the pages at the root
+    /// of docs/website, which have no section to sort them: introduction.md declares a
+    /// 'sidebar_position' and badge.md does not, and that is exactly the split between a link that
+    /// belongs in the main body and one that belongs under "Optional".
+    /// </param>
+    sealed record DocPage(
+        string Title,
+        string Description,
+        string Url,
+        string Section,
+        int SectionOrder,
+        int Order,
+        bool IsPrimary);
+
+    // Only the three keys this index reads. Docusaurus accepts many more (tags, slug, keywords,
+    // hide_title...), so unmatched ones are ignored rather than failing a page that adds one.
+    sealed record Frontmatter
     {
-        var match = OrderPrefix.Match(segment);
-        return match.Success ? int.Parse(match.Groups["order"].Value) : int.MaxValue;
+        public string Title { get; init; }
+
+        public string Description { get; init; }
+
+        public int? SidebarPosition { get; init; }
+    }
+
+    Target GenerateLlmsTxt => _ => _
+        .Executes(() =>
+        {
+            var pages = ReadDocPages();
+            LlmsTxtFile.WriteAllText(RenderLlmsTxt(pages));
+
+            Log.Information("Wrote {File} with {Count} pages", RootDirectory.GetUnixRelativePathTo(LlmsTxtFile), pages.Count);
+        });
+
+    // CI gate, in the shape of VerifyGeneratedTools: GenerateLlmsTxt only runs when a contributor
+    // remembers to invoke it, so a page added under docs/website without regenerating would merge
+    // with docs/llms.txt silently missing it. `Requires` is asserted for the whole scheduled plan
+    // before any target runs, so the "start clean" check below still fires before GenerateLlmsTxt
+    // regenerates anything; the explicit re-check afterward catches drift with a message pointing
+    // at the fix.
+    //
+    // Wired into BOTH workflows on purpose. build.yml ignores docs/**, so on its own it would never
+    // fire on the change that actually invalidates the file; build-skip.yml is the workflow that
+    // handles those PRs, and it runs this target for exactly that reason.
+    Target VerifyLlmsTxt => _ => _
+        .Requires(() => GitHasCleanWorkingCopy())
+        .DependsOn(GenerateLlmsTxt)
+        .Executes(() =>
+        {
+            Assert.True(
+                GitHasCleanWorkingCopy(),
+                "docs/llms.txt is out of sync with docs/website. Run './build.ps1 GenerateLlmsTxt' "
+                + "locally and commit the result.");
+        });
+
+    // Docusaurus routes both .md and .mdx, and excludes anything whose file or directory name
+    // starts with an underscore (**/_*.md, **/_*/**). docs/website/_snippets/ exists for exactly
+    // that reason, so indexing it would emit URLs the site never serves.
+    IReadOnlyList<DocPage> ReadDocPages()
+    {
+        return DocsWebsiteDirectory.GlobFiles("**/*.md", "**/*.mdx")
+            .Where(x => !DocsWebsiteDirectory.GetUnixRelativePathTo(x).ToString()
+                .Split('/')
+                .Any(segment => segment.StartsWith('_')))
+            .Select(ReadPage)
+            .OrderBy(x => x.SectionOrder)
+            .ThenBy(x => x.Order)
+            // Ordinal, not the culture-sensitive default: docs/llms.txt is verified byte for byte,
+            // so a contributor on another culture must not regenerate a differently ordered file
+            // and trip VerifyLlmsTxt with no real drift.
+            .ThenBy(x => x.Title, StringComparer.Ordinal)
+            .ToList();
+    }
+
+    DocPage ReadPage(AbsolutePath file)
+    {
+        var lines = file.ReadAllLines();
+        var frontmatterEnd = GetFrontmatterEnd(lines);
+        var frontmatter = ReadFrontmatter(lines, frontmatterEnd);
+
+        // Docusaurus falls back to the first H1 when a page declares no 'title', and docs/website
+        // has a page that relies on it: badge.md carries no frontmatter at all and is served as
+        // "Badge". Rejecting it would refuse a page the site renders correctly, so the fallback
+        // matches Docusaurus. A page with neither still fails, because that leaves no link text.
+        var title = frontmatter.Title.IsNullOrWhiteSpace()
+            ? GetFirstHeading(lines, frontmatterEnd)
+            : frontmatter.Title;
+        Assert.NotNullOrWhiteSpace(
+            title,
+            $"{DocsWebsiteDirectory.GetUnixRelativePathTo(file)} has neither a 'title' in its "
+            + "frontmatter nor a top-level heading. One of the two is needed: it is the link text "
+            + "in docs/llms.txt.");
+
+        var description = frontmatter.Description.IsNullOrWhiteSpace()
+            ? GetFirstProseParagraph(lines, frontmatterEnd)
+            : frontmatter.Description;
+
+        var relative = DocsWebsiteDirectory.GetUnixRelativePathTo(file).ToString();
+        var segments = relative.Split('/');
+        var isNested = segments.Length > 1;
+
+        return new DocPage(
+            Title: title,
+            Description: Summarize(description),
+            Url: ToPublicUrl(file),
+            // Root-level pages have no section, so they are rendered either above the first one or
+            // under "Optional", depending on IsPrimary.
+            Section: isNested ? ToSectionTitle(segments[0]) : null,
+            SectionOrder: isNested ? GetOrder(segments[0]) : int.MaxValue,
+            // Docusaurus lets a page's own 'sidebar_position' override the numeric filename prefix,
+            // and docs/website uses it: 07-ide has no prefixes, and rider.md declares position 1 to
+            // sort first. Reading only the prefix would order that section by title instead.
+            Order: frontmatter.SidebarPosition ?? GetOrder(segments[^1]),
+            IsPrimary: isNested || frontmatter.SidebarPosition.HasValue);
     }
 
     string ToPublicUrl(AbsolutePath page)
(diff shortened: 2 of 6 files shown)


.github/workflows/build-skip.yml
```diff
@@ -9,17 +9,29 @@
...
 # a status check named `ubuntu-latest`; the workflow `name:` mirrors build.yml's
 # `build` so docs-only runs appear under the same workflow in the Actions tab.
 
 name: build
 
+# Mirrors build.yml. This job does real work now, so rapid pushes to a docs PR
+# would otherwise stack concurrent builds, which is the cost this file exists to
+# avoid in the first place.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
dennisdoomen: 🔧🤖 Same concurrency group as build.yml (both `name: build`) — a PR touching code + a `.md` file triggers both workflows, and `cancel-in-progress` lets one cancel the other.

build/Build.Documentation.cs
```diff
@@ -0,0 +1,353 @@
...
+
+        var end = Array.FindIndex(lines, startIndex: 1, x => x.Trim() == "---");
+        // An opened but unclosed block is malformed. Returning 0 would hand the delimiter and the
+        // key/value lines to the prose reader and ship them as a description, so fail instead: a
+        // wrong entry in a generated index is worse than a build that says what is wrong.
+        Assert.True(end >= 0, "Frontmatter is opened with '---' but never closed.");
+        return end + 1;
+    }
+
+    static Dictionary<string, string> ReadFrontmatter(string[] lines, int frontmatterEnd)
```
dennisdoomen: ♻️🤖 This hand-rolled frontmatter parser duplicates `Fallout.Utilities.Text.Yaml.YamlExtensions.GetYaml<T>` (with `UnderscoredNamingConvention`), which already exists in this repo.

build/Build.Documentation.cs
```diff
@@ -0,0 +1,353 @@
...
+        {
+            var trimmed = line.Trim();
+
+            if (trimmed.StartsWith("```"))
+            {
+                insideFence = !insideFence;
+                continue;
+            }
+
+            var isProse = !insideFence &&
```
dennisdoomen, line 264: 🤔🤖 Excludes headings/tables/images/imports but not list items (`-`/`*`) or blockquotes (`>`) — a future page opening with a list would garble the description.

build/Build.Documentation.cs
```diff
@@ -0,0 +1,353 @@
...
+        // key/value lines to the prose reader and ship them as a description, so fail instead: a
+        // wrong entry in a generated index is worse than a build that says what is wrong.
+        Assert.True(end >= 0, "Frontmatter is opened with '---' but never closed.");
+        return end + 1;
+    }
+
+    static Dictionary<string, string> ReadFrontmatter(string[] lines, int frontmatterEnd)
+    {
+        var entries = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+        for (var i = 1; i < Math.Max(frontmatterEnd - 1, 1); i++)
```
dennisdoomen: ⛏🤖 `Math.Max(frontmatterEnd - 1, 1)` is a no-op: `GetFrontmatterEnd` only ever returns 0 or ≥2, so the clamp to 1 never changes the loop bound.

build/Build.Documentation.cs
```diff
@@ -0,0 +1,353 @@
...
+    // https://docs.fallout.build/sitemap.xml, not against the README: the README's own links omit
+    // the prefix and 404 (see the follow-up on the PR). Hardcoded for the same reason as
+    // CanonicalRepositoryIdentifier: the generated file must be identical whichever fork
+    // regenerates it.
+    const string DocsBaseUrl = "https://docs.fallout.build/docs/";
+
+    // Docusaurus orders pages by a numeric prefix on the directory and file name, and strips that
+    // prefix from the served URL. So 01-getting-started/01-installation.md is served at
+    // /docs/getting-started/installation.
+    static readonly Regex OrderPrefix = new(@"^(?<order>\d+)-", RegexOptions.Compiled);
```
dennisdoomen: 🔧🤖 `OrderPrefix`/`InlineLink`/`FrontmatterEntry` (private static readonly, non-const) should be camelCase per .editorconfig's naming rule, like `unicodeRegex` elsewhere in src/.

build/Build.Documentation.cs
```diff
@@ -0,0 +1,353 @@
...
+    // fire on the change that actually invalidates the file; build-skip.yml is the workflow that
+    // handles those PRs, and it runs this target for exactly that reason.
+    Target VerifyLlmsTxt => _ => _
+        .Requires(() => GitHasCleanWorkingCopy())
+        .DependsOn(GenerateLlmsTxt)
+        .Executes(() =>
+        {
+            Assert.True(
+                GitHasCleanWorkingCopy(),
+                "docs/llms.txt is out of sync with docs/website. Run './build.ps1 GenerateLlmsTxt' locally and commit the result.");
```
dennisdoomen: ⛏🤖 Line is 132 chars, over .editorconfig's `max_line_length = 130`.

build/Build.Documentation.cs
```diff
@@ -0,0 +1,353 @@
...
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using Fallout.Common;
+using Fallout.Common.IO;
+using Fallout.Common.Utilities;
+using Fallout.Common.Utilities.Collections;
+using Serilog;
+using static Fallout.Common.Tools.Git.GitTasks;
+
```
dennisdoomen, line 16: 🔧🤖 Member order is bottom-up (helpers → entry points) and fields are split into two clusters; `RewriteCsprojsStep.cs` shows this repo's convention is fields-grouped, then caller-before-callee.

dennisdoomen (commented):
Automated review pass (correctness, design, and style).

.github/workflows/build-skip.yml
phmatray: Fixed in 79b65f5c. The group is now the literal `build-skip-${{ github.ref }}`, with a comment recording why it cannot be `${{ github.workflow }}`: that expands to the workflow `name:`, which is deliberately `build` in both files so docs-only runs report under the same status check.

build/Build.Documentation.cs
phmatray: Done in 4929f86a. `ReadFrontmatter` now deserializes onto a typed `Frontmatter` record via `GetYaml<T>`, and the hand-rolled line regex is gone.

One thing your comment did not mention but that the swap needs: the builder also has to be `.IgnoreUnmatchedProperties()`. YamlDotNet throws on unmatched keys by default, and Docusaurus accepts plenty this index does not read (`tags`, `slug`, `keywords`, `hide_title`), so without it the first page to add one would fail the build.

`docs/llms.txt` regenerates byte for byte identical, so the change moves nothing in the generated index.

build/Build.Documentation.cs
phmatray: Done in 4929f86a. `- `, `* ` and `>` now join the exclusions. The two list markers are matched with the trailing space on purpose, so a `---` rule or a `**bold**` lead-in is still read as the prose it is.

build/Build.Documentation.cs
phmatray: Correct — dropped in 4929f86a. `GetFrontmatterEnd` returns 0 or at least 2, and at 0 the loop never entered with or without the clamp.

build/Build.Documentation.cs
phmatray: Done in 4929f86a: `orderPrefix` and `inlineLink`. `FrontmatterEntry` went away entirely with the parser swap above.

Checked against .editorconfig first — `private_fields_should_be_camel_case` is severity `warning` and applies to fields without a `const` modifier, and `src/` has no PascalCase counter-example.

build/Build.Documentation.cs
phmatray: Done in 4929f86a — wrapped onto two concatenated strings. It was the only line over 130 in the file.

build/Build.Documentation.cs
phmatray: Done in 4929f86a: fields grouped first, then the two records, then the targets, then the helpers caller-before-callee.

Modelled on `build/Build.CodeGeneration.cs` rather than `RewriteCsprojsStep.cs` — it is the closer analogue (same `Generate`/`Verify` target pair, same file layer) and already reads fields-then-targets.

The reorder moved whole member blocks with their comments attached, asserting the file's line multiset was unchanged, so no code was retyped in the process.

dennisdoomen (approved):

dennisdoomen (approved):

dennisdoomen:
@phmatray please fix the conflicts and then we can move forwrd

phmatray:
## Conflict with `develop` resolved

Merged `upstream/develop` into the branch (`595848e1`). One file conflicted: `build/Build.CI.GitHubActions.cs`.

**Cause.** `develop` ran a Rider code cleanup that rewrote the `[GitHubActions]` attribute arrays from one line to one item per line. This branch had edited the same single-line `InvokedTargets`, so the two edits overlapped.

**Resolution.** Kept `develop`'s multi-line layout and put `nameof(VerifyLlmsTxt)` back into the `build` workflow's `InvokedTargets`, after `VerifyGeneratedTools`:

```csharp
InvokedTargets = new[]
{
    nameof(VerifyGeneratedTools),
    nameof(VerifyLlmsTxt),
    nameof(ITest.Test),
    nameof(IPack.Pack)
},
```

`develop`'s side of the conflict was formatting only. No behaviour changed, and the generated `.github/workflows/*.yml` and `.fallout/build.schema.json` are unchanged by the merge.

**Verified after merging:**

- `./build.sh VerifyGeneratedTools VerifyLlmsTxt` exits 0. This is the same gate pair CI runs, minus `Test` and `Pack`.
- The build project still compiles against `develop`'s changes under `src/`.
- Regeneration is a no-op: `[INF] Wrote docs/llms.txt with 37 pages` and `git status` stays clean, so `docs/llms.txt` is byte-identical to the committed file.
- `.github/workflows/build.yml:59` still reads `dotnet fallout VerifyGeneratedTools VerifyLlmsTxt Test Pack`.

The PR is mergeable again.

ChrisonSimtian (approved):

ChrisonSimtian:
@phmatray please go ahead and merge this at your earliest convenience :-) and then please ping me for the other half of the PR once its up

phmatray:
Hi @ChrisonSimtian ,
I don't have permission to merge into this repository.
If you have the rights, please go ahead :)

I'll tackle the second PR this week.

ChrisonSimtian:
> Hi @ChrisonSimtian ,
> 
> I don't have permission to merge into this repository.
> 
> If you have the rights, please go ahead :)
> 
> 
> 
> I'll tackle the second PR this week.

dennis beat me to it 😬 thanks for sorting this one out

Issue #644, opened 2026-08-18
Declare target effects and gate them at plan time


ChrisonSimtian:
### Problem

Fallout targets range from writing into `artifacts/` to publishing packages to nuget.org and tagging releases. Nothing in the target model distinguishes them. A caller that wants to run a build without changing anything outside the working tree has no way to ask which targets are safe.

For an automated caller the only safe policy is therefore to never run the build unattended, which removes most of the value. The Continuous Delivery direction (milestone #8) adds deployment targets, which widens the gap.

### Outcome

A target declares the class of effect it has, and the runner refuses to reach a target whose effect class was not granted.

Sketch:

```csharp
Target Compile => _ => _
    .Effect(Effects.LocalFilesystem)
    .Executes(...);

Target PushToNuGet => _ => _
    .Effect(Effects.RemoteMutation, Effects.Irreversible)
    .Executes(...);
```

`fallout run Pack --allow local` succeeds. A plan that reaches a `RemoteMutation` target fails during planning, before any target executes, naming the target and the grant it needs.

### Acceptance criteria

- [ ] Targets can declare one or more effect classes
- [ ] Effect classes are checked at plan time, so a run stops before the first target executes
- [ ] The error names the target and the grant required
- [ ] Effect classes propagate through dependencies: depending on a remote-mutation target requires that grant
- [ ] `describe --json` exposes each target's effect classes

### Open questions

- Is an undeclared target permissive or restrictive by default? Restrictive is safer, but changes behaviour for every existing build.
- Fixed set of effect classes, or an open set that plugins can extend?
- Does the contract belong in the plugin SDK, given #100 (plugin discovery and load model) has to settle a related trust question?

### Notes

Useful outside automation too: `--allow remote` as a CI-only flag prevents accidental publishes from a developer machine.

ChrisonSimtian:
Part of #648 (the AI-native build interface epic), milestone [AI-native build interface](https://github.com/Fallout-build/Fallout/milestone/24).

Issue #312, opened 2026-05-29
[Foundation] FT-7: Formalize the build extension pipeline (deterministic ordering + typed phases)


ChrisonSimtian:
## Summary

Formalize the build extension/middleware pipeline: deterministic ordering and a typed phase model instead of comment-encoded magic priority numbers. Directly supports the plugin architecture foundation in this milestone.

## Findings (file:line)

- `FalloutBuild.Events.cs:21-25` — extensions discovered via `GetCustomAttributes<BuildExtensionAttributeBase>()` and ordered `OrderByDescending(x => x.Priority)`. **Equal priorities fall back to CLR reflection order → non-deterministic across runtimes.**
- `FalloutBuild.cs:44-62` — ~15 built-in extensions share magic priorities (150/100/75/50/10/5/-100); the "before logo / after logo / after finish" phases exist only as comments.
- `Execution/Extensibility/BuildExtensionAttributeBase.cs` — `IBuildExtension` + `IOn*` lifecycle interfaces are public; concrete extensions are mostly internal (keep that).

## Scope

- Add a deterministic secondary ordering key for equal priorities.
- Introduce a typed phase model (enum or ordered phase tokens) replacing comment-encoded ranges.
- Document the `IOn*` lifecycle contract.
- Confirm middleware internals stay `internal` (milestone policy: no public plugin SDK yet).

## Acceptance criteria

- Equal-priority ordering is deterministic and covered by a test.
- Phases are typed, not magic numbers.
- No public plugin surface added.

## Notes

- **Non-breaking.**
- Depends on: FT-2 (can proceed largely in parallel with FT-3/4/5).
- Size: **M**.

PR #662, opened 2026-08-27
Expose the build model as JSON via --describe and --plan --json (#642)

### LINKED ISSUE #499: Emit declared build parameters into build-graph.json
**Problem**

`build-graph.json` (the editor-tooling contract in `BuildGraphUtility`) carries only targets. The build's `[Parameter]`/`[Secret]` members — e.g. `NuGetApiKey` on `IPublish`, the set `dotnet fallout --help` lists — never reach the VS Code extension, so it can't surface them.

**Outcome**

Additive `parameters[]` in `build-graph.json`, projected from the build's declared parameters. Schema stays **v1** (new field is backward-compatible):

```jsonc
{ "name": "NuGetApiKey", "secret": true, "description": "…", "declaredIn": "IPublish" }
```

**Acceptance criteria**

- All declared `[Parameter]` members are projected, **including inherited component params** (IPublish, etc.) — matching `--help`'s set.
- `[Secret]` members flagged `secret: true`.
- Fields: `name`, `secret`, `description`, `declaredIn`.
- `SchemaVersion` unchanged (additive); existing consumers unaffected.
- `BuildGraphUtilitySpecs` updated to cover the projection.

Context: [ADR-0009] run-config discussion; feeds the extension issue below.

### LINKED ISSUE #616: `--help` crashes with IOException when stdout has no console handle
@
## Problem

`HandleHelpRequestsAttribute` reads `Console.BufferWidth` directly, so any `--help` invocation where the process has no console buffer (redirected stdout, pipe, CI, IDE output pane) throws instead of printing help. The target list prints, then the run hard-fails.

`src/Fallout.Build/Execution/Extensions/HandleHelpRequestsAttribute.cs:62`:

```csharp
if (nextLength >= Console.BufferWidth || nextLength > 90)
```

Repro — migrated consumer repo (FluentAssertions at its last NUKE commit), `Fallout.Common` 10.4.0-rc.5:

```
./build.ps1 --help > log.txt 2>&1
```

```
System.IO.IOException: The handle is invalid.
   at System.ConsolePal.GetBufferInfo(Boolean throwOnNoConsole, Boolean& succeeded)
   at System.Console.get_BufferWidth()
   at HandleHelpRequestsAttribute...SplitLines|3(String text)  HandleHelpRequestsAttribute.cs:line 62
   at HandleHelpRequestsAttribute.GetParametersText()          HandleHelpRequestsAttribute.cs:line 90
   at HandleHelpRequestsAttribute.OnBuildInitialized(...)      HandleHelpRequestsAttribute.cs:line 20
   at BuildManager.Execute[T](...)                             BuildManager.cs:line 68
```

Affects **both `main` and `release/v10.4`** — the call site is identical on each.

Note the codebase already has the right abstraction (`ConsoleUtility` → `ConsoleWrapper.BufferWidth`), but `SystemConsole.BufferWidth` also calls `Console.BufferWidth` unguarded — so switching to the wrapper alone does not fix this.

## Outcome

`--help` prints usable help regardless of whether a console is attached.

## Acceptance criteria

- [ ] `--help` succeeds with stdout redirected to a file and through a pipe
- [ ] Width falls back to a sane default (the existing `90` clamp) when no console buffer is available
- [ ] `SystemConsole.BufferWidth` guards on `Console.IsOutputRedirected` rather than throwing
- [ ] Regression test covers the redirected-stdout path
- [ ] Backported to `release/v10.4`
@

### LINKED ISSUE #642: Expose the build model as data: fallout describe --json and plan --json
### Problem

To learn what a Fallout build can do, a tool has to read `Build.cs`, follow its partial classes, resolve the attributes, and re-run the dependency resolution in its head. `--help` renders that same information as human text, and there is no machine-readable form of it. There is also no way to ask what a target *would* run without running it.

### Outcome

Two read-only commands that print the build model as JSON and never execute a target.

- `fallout describe --json` — every target, the dependency edges (`DependsOn`, `Before`, `After`, `Triggers`), every parameter with its CLR type, default value and required flag, tool requirements, and descriptions.
- `fallout plan <target> --json` — the ordered list of targets that would run, with skip reasons. This is the existing plan resolution with the executor not attached.

### Acceptance criteria

- [ ] `fallout describe --json` emits a schema-versioned document covering targets, dependency edges, parameters, and tool requirements
- [ ] `fallout plan <target> --json` emits the resolved plan and invokes no target
- [ ] Both commands leave the working tree untouched and run no external tool
- [ ] Both exit non-zero with a machine-readable error when the build project fails to load
- [ ] `--help` is rendered from the same model, so the human and machine views cannot drift

### Notes

`src/Fallout.Build/Execution/ExecutionPlanner.cs` already computes the plan, then discards it into the console. #391 (the Serilog and NDJSON output work) names `describe --json` as a motivation but does not track it.

phmatray:
Implements #642.

Closes #642.

Executing the implementation plan task-by-task; the checklist below — and the plan on the issue — are ticked as each task lands. Opened as a draft — will be marked ready after the final task and a code-review pass.

Two constraints shaped the design, both verified in the tree:

- `BuildGraphUtility` already projects targets and all four relation kinds into a schema-versioned document, so this extends that model additively rather than inventing a second one. `SchemaVersion` stays `1`, which also delivers #499's `parameters[]`.
- `BuildManager.Execute` runs `ToolRequirementService.EnsureToolRequirements` (writes `nuget.csproj`, shells out to `dotnet restore`) **before** the `IOnBuildInitialized` hooks. So the "runs no external tool" criterion rules out another extension; the emission short-circuits above that line.

The commands land as the build-side parameters `--describe` and `--plan --json`, because `CommandDispatcher` routes bare arguments to the build and reserves `:`-prefixed tokens for CLI commands.

### Plan
- [x] Task 1: Project tool requirements onto the target model
- [x] Task 2: Project declared parameters onto the model (also closes #499)
- [x] Task 3: `--describe` emits the model before tool requirements run
- [x] Task 4: `--plan --json` emits the resolved execution plan
- [x] Task 5: Machine-readable error envelope on failure
- [x] Task 6: Render `--help` from the shared model

### `--help` is byte-identical to `main`

Verified by building `main` in a scratch worktree and diffing the output: the only difference is the version string. `--help` takes its displayed *fields* from the shared model (so the two views cannot drift) but keeps *ordering* — of targets, of each dependency line, and of parameters — from the declarations, because declaration order carries the pipeline reading a human wants.

### Code review

A review of the branch surfaced 15 findings; all are addressed. The ones worth calling out:

- **The plan contradicted the executor.** `BuildExecutor.MarkTargetSkipped` only skips when `!target.Invoked`, so `--skip` never stops an explicitly invoked target. The projection had no such guard and reported `"skip": "via parameter"` for it — a consumer driving CI off the plan would conclude nothing runs. Fixed, with a spec that pins the parity.
- **The gate was in the wrong place.** It sat after the `IOnBuildCreated` extensions, so each misbehaving extension needed its own opt-out — and `UpdateNotificationAttribute` calls `Console.ReadKey()`, which would have deadlocked a piped `--describe`. The gate now fires before any extension runs, and the per-extension opt-out is gone. `IsRequested` reads each flag from the injected property *or* the raw arguments, because `InjectParameterValuesAttribute` is itself an `IOnBuildCreated` extension.
- **Build-level `[Requires<T>]` was invisible.** That attribute targets Class/Interface only, so class-level requirements could never appear on a target and were missing entirely. They now project to a root-level `toolRequirements[]`.
- **`Type.FullName` leaked the runtime version** into the contract for constructed generics (`List\`1[[System.String, …, Version=10.0.0.0, …]]`). Generic arguments are now rendered recursively.
- Dropped a `list` field that could never be false, gave the plan and error documents their own schema-version constants, and de-duplicated the serializer options, version lookup, and skip-reason string that had been copy-pasted.

### Making standard output actually parseable

Running the feature revealed that the document was correct but **unreachable through any shipped entry point** — everything upstream of the build wrote to stdout too, so `--describe` could not be piped into a parser. Three fixes:

- **`Fallout.Cli`** compiled the build project with its output inherited on stdout. For an introspection request that step now goes to standard error. Measured through the locally built CLI: 12 123 bytes of JSON on stdout, 8 582 bytes of build noise on stderr. Ordinary runs are unchanged — build output stays on stdout.
- **`build.ps1` / `build.sh`** (and the shipped templates, kept in sync) printed their banner and `dotnet tool restore` output to stdout. That is provisioning diagnostics, so it moves to standard error.
- **`--help` no longer aborts when stdout has no console.** `Console.BufferWidth` throws `IOException` behind a pipe or on a console-less agent, which killed `--help` after the target list and before the parameters. The wrap width is cosmetic, so it now falls back to the existing 90-column cap. **This closes #616.**

Also: the emitted documents are newline-terminated, matching `SchemaUtility`'s JSON.

⚠️ `./build.ps1 --describe` stays unparseable until a Fallout release carrying the CLI fix is installed, because the script delegates to whatever `dotnet fallout` is on the machine. Nothing in this PR can change an already-installed tool.

### Follow-ups

- A versionless `[Requires]` resolves its version over the network inside `NuGetPackageRequirement`'s constructor, before any short-circuit. Pre-existing, affects `--help` too; "runs no external tool" is scoped to this code path.
- The `IOnBuildPlanned` extension hook this deliberately avoided is worth revisiting once FT-7 (#312) formalizes the extension pipeline; #644 would use it too.

### DIFF
--- a/.fallout/build.schema.json
+++ b/.fallout/build.schema.json
@@ -62,6 +62,10 @@
           "type": "boolean",
           "description": "Indicates to continue a previously failed build attempt"
         },
+        "Describe": {
+          "type": "boolean",
+          "description": "Prints the build model as JSON and exits without executing any target"
+        },
         "Help": {
           "type": "boolean",
           "description": "Shows the help text for this build assembly"
@@ -69,6 +73,10 @@
         "Host": {
           "description": "Host for execution. Default is 'automatic'",
           "$ref": "#/definitions/Host"
+        },
+        "Json": {
+          "type": "boolean",
+          "description": "Emits machine-readable JSON on standard output for read-only requests (--plan)"
         },
         "NoLogo": {
           "type": "boolean",
--- a/build.ps1
+++ b/build.ps1
@@ -4,7 +4,9 @@ Param(
     [string[]]$BuildArguments
 )
 
-Write-Output "PowerShell $($PSVersionTable.PSEdition) version $($PSVersionTable.PSVersion)"
+# Provisioning chatter goes to standard error, so standard output carries only what the build
+# itself writes. That is what lets `./build.ps1 --describe | ConvertFrom-Json` work.
+[Console]::Error.WriteLine("PowerShell $($PSVersionTable.PSEdition) version $($PSVersionTable.PSVersion)")
 
 Set-StrictMode -Version 2.0; $ErrorActionPreference = "Stop"; $ConfirmPreference = "None"; trap { Write-Error $_ -ErrorAction Continue; exit 1 }
 $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent
@@ -60,7 +62,8 @@ else {
     $env:PATH = "$DotNetDirectory;$env:PATH"
 }
 
-Write-Output "Microsoft (R) .NET SDK version $(& $env:DOTNET_EXE --version)"
+[Console]::Error.WriteLine("Microsoft (R) .NET SDK version $(& $env:DOTNET_EXE --version)")
 
-ExecSafe { & $env:DOTNET_EXE tool restore }
+# PowerShell has no `1>&2`, so the restore's output is pumped to standard error explicitly.
+ExecSafe { & $env:DOTNET_EXE tool restore 2>&1 | ForEach-Object { [Console]::Error.WriteLine($_) } }
 ExecSafe { & $env:DOTNET_EXE fallout $BuildArguments }
--- a/build.sh
+++ b/build.sh
@@ -53,7 +53,9 @@ else
     export PATH="$DOTNET_DIRECTORY:$PATH"
 fi
 
-echo "Microsoft (R) .NET SDK version $("$DOTNET_EXE" --version)"
+# Provisioning chatter goes to standard error, so standard output carries only what the build
+# itself writes. That is what lets `./build.sh --describe | jq` work.
+echo "Microsoft (R) .NET SDK version $("$DOTNET_EXE" --version)" >&2
 
-"$DOTNET_EXE" tool restore
+"$DOTNET_EXE" tool restore >&2
 exec "$DOTNET_EXE" fallout "$@"
--- a/src/Fallout.Build/Execution/BuildExecutor.cs
+++ b/src/Fallout.Build/Execution/BuildExecutor.cs
@@ -15,6 +15,13 @@ namespace Fallout.Common.Execution;
 /// </summary>
 internal static class BuildExecutor
 {
+    /// <summary>
+    /// Reason recorded when <c>--skip</c> names a target. Shared with
+    /// <see cref="BuildIntrospectionService" /> so the predicted plan and the executed one describe
+    /// a skip identically.
+    /// </summary>
+    internal const string SkippedViaParameterReason = "via parameter";
+
     // NOTE: no IFalloutBuild because of BuildAttemptFile + WriteTarget
     private static AbsolutePath BuildAttemptFile => Constants.GetBuildAttemptFile(FalloutBuild.RootDirectory);
 
@@ -25,7 +32,7 @@ internal static class BuildExecutor
             skippedTargets = skippedTargets.Select(x => x.Replace("-", string.Empty)).ToArray();
             build.ExecutionPlan
                 .Where(x => skippedTargets.Count == 0 || skippedTargets.Contains(x.Name, StringComparer.OrdinalIgnoreCase))
-                .ForEach(x => MarkTargetSkipped(build, x, reason: "via parameter"));
+                .ForEach(x => MarkTargetSkipped(build, x, reason: SkippedViaParameterReason));
         }
 
         build.ExecutionPlan.ForEach(x => CheckConditions(build, x, x.StaticConditions));
--- a/src/Fallout.Build/Execution/BuildIntrospectionService.cs
+++ b/src/Fallout.Build/Execution/BuildIntrospectionService.cs
@@ -48,6 +48,22 @@ internal static class BuildIntrospectionService
     private static bool Flag(bool injected, string parameterName)
         => injected || ParameterService.GetParameter<bool>(parameterName);
 
+    /// <summary>
+    /// Whether raw command-line arguments request introspection, for callers that must decide
+    /// before a build process exists — the CLI, which has to know where to send the build step's
+    /// own output. Shares this type with the property-based overload so the two entry points
+    /// cannot disagree about what counts as a read-only request.
+    /// </summary>
+    internal static bool IsRequested(IReadOnlyCollection<string> arguments)
+        => HasFlag(arguments, nameof(FalloutBuild.Describe)) ||
+           (HasFlag(arguments, nameof(FalloutBuild.Plan)) && HasFlag(arguments, nameof(FalloutBuild.Json)));
+
+    // Accepts every spelling the parameter parser does: --describe, -describe, --DESCRIBE.
+    private static bool HasFlag(IReadOnlyCollection<string> arguments, string parameterName)
+        => arguments.Any(x =>
+            x.StartsWith("-", StringComparison.Ordinal) &&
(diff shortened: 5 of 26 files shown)


ChrisonSimtian (changes requested):
a lot of this code is static methods and static classes. I think this one needs a bit more hands on from a human, the code quality unfortunately trips AI to think all-static is a good and preferred pattern.
might be worth to nudge your agent into the right direction and have it have another go 😊

phmatray:
Fair call, and it turned out to be hiding a real bug rather than just style. Pushed in e6c358d5.

`BuildIntrospectionService` is a sealed class now, holding the resolved request. `BuildManager` builds one with `For(build)` and reuses it at the gate, on the failure path, and before the outcome tables. Those three sites each re-read the flags out of `ParameterService`, so a single run answered "is this an introspection request?" three separate times from process-global state that nothing kept in agreement. The invoked targets are read once now too, rather than once for the planner and again inside `GetDocument`.

The part worth flagging: dropping the static shape removed a test-only overload, and that overload was papering over a coverage gap. Production described a build through `GetJsonString(build, targets)`, which projects the class-level `[Requires<T>]` requirements. The specs called `GetDescribeJson(build, targets, version)`, which routed to a `GetModel` overload passing **no** requirements — so the specs were asserting a document the product never emits, and that projection had no coverage at all. `BuildGraphUtility` now takes the version on the same path production uses, and there is a new spec pinning a build-level requirement reaching the document.

One thing I deliberately left static: `IsRequested(IReadOnlyCollection<string>)`. The CLI calls it before a build process exists, and it is a pure function of its arguments, so an instance would only add ceremony. Happy to move it if you would rather it lived somewhere else.

I did not touch `BuildGraphUtility`'s other members or `BuildManager`/`BuildExecutor` themselves — they are pre-existing static utilities and pure projections, and converting them looked like a bigger change than this PR should carry. Say the word if you want that too.

Full suite green: 849 passing.

phmatray:
## Conflict with `develop` resolved

Merged `upstream/develop` into the branch (`cb0bdc22`). Four files conflicted, all under `src/Fallout.Build/Execution/`.

**Cause.** `develop` ran a formatting pass over that folder: braces around single-statement blocks, blank lines between statements, re-indented array initializers, and one `using` moved. On `develop`'s side every conflicting change was style only. This branch had rewritten the same regions for real.

**Resolution.** Kept this branch's logic in all four files and applied `develop`'s style around it.

| File | What was kept |
|---|---|
| `BuildManager.cs` | `Finish()` keeps the `introspection is { IsRequestedForRun: true }` guard, now braced. |
| `Extensions/HandleHelpRequestsAttribute.cs` | The projected model, the `GetBufferWidth()` fallback and the `(Member, Model)` tuple. Braces and blank lines follow `develop`. |
| `Extensions/SerializeBuildGraphAttribute.cs` | Both `using`s, plus `develop`'s unqualified `Log.Verbose`. |
| `ValueInjectionUtility.cs` | The `Describe` and `Json` entries, inside `develop`'s re-indented array. |

One change sits just outside a conflict hunk: the `if (!projected.Listed)` guard in `GetTargetsText` is now braced. Leaving it alone would have made that guard the only unformatted statement in a file `develop` had just formatted end to end.

**Verified after merging:**

- `dotnet build fallout.slnx` — 0 errors.
- `dotnet test fallout.slnx` — 0 failures across all 13 test projects, including `BuildGraphUtilitySpecs`, `BuildIntrospectionServiceSpecs` and `HandleHelpRequestsSpecs`. No Verify snapshot drift, so no `.received.` file was produced.
- `./build.sh --help` renders `--describe` and `--json` through the merged `HandleHelpRequestsAttribute`, which was the largest conflict.
- `./build.sh --describe` emits a well-formed document (19 targets, 22 parameters) and `./build.sh --plan --json` emits `version` / `invokedTargets` / `plan`. Both exercise the `BuildManager` short-circuit and the `Finish()` guard that were resolved by hand.
- `.fallout/build.schema.json` does not drift: booting the build project leaves `git status` clean.

The PR is mergeable again.

https://claude.ai/code/session_01EnT6hhAWLHoiVdx8NMAu7u

PR #671, opened 2026-09-10
Fix the `:add-package` command by not using variable parsing from `.\build.ps1`

### LINKED ISSUE #670: .\build.ps1 in its current state isn't capable for :add-package command
### Usage Information

Fallout 10.4.0.15 (Global Tool)

### Description

Running `fallout :add-package Package --version x.y.z` does not work anymore.

The problem is the fancy "Join-Path ..." in the ps1 bootstrapper script.
This causes the line 35-36 of `Fallout.Cli.AddPackageCommand` to give this as build script path: `Join-Path C:\dev\RootDir 'build/_build.csproj'` which is utterly wrong...

### Reproduction Steps

- Run `fallout :add-package GitVersion.Tool --version 6.8.2` on a solution with Fallout installed
- (needs the line `$BuildProjectFile = Join-Path "$PSScriptRoot" 'build/_build.csproj'` in build.ps1, otherwise it fails anyway with a "KeyNotFoundException)

### Expected Behavior

Adds the package

### Actual Behavior

Gives a powershell only command back

### Regression?

Yes. Works on solutions which were set up with NUKE, which have this instead of "Join-Path": `$BuildProjectFile = "$PSScriptRoot\build\_build.csproj"`

### Known Workarounds

Restoring this exact string

Yes

ITaluone:
The `Join-Path` actually prevents to use the gloabl tool's `:add-package` command, since this command greps the build.csproj path from the `.\build.ps1`. `./build.sh` is not affected, since there is no such "fancy" command to "just" join paths :).

Fixes: #670

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -13,9 +13,9 @@ $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent
 # CONFIGURATION
 ###########################################################################
 
-$TempDirectory = Join-Path $PSScriptRoot '.fallout/temp'
+$TempDirectory = "$PSScriptRoot\\.fallout\temp"
 
-$DotNetGlobalFile = Join-Path $PSScriptRoot 'global.json'
+$DotNetGlobalFile = "$PSScriptRoot\global.json"
 $DotNetInstallUrl = "https://dot.net/v1/dotnet-install.ps1"
 $DotNetChannel = "STS"
 
--- a/build.ps1
+++ b/build.ps1
@@ -13,9 +13,9 @@ $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent
 # CONFIGURATION
 ###########################################################################
 
-$TempDirectory = "$PSScriptRoot\\.fallout\temp"
+$TempDirectory = "$PSScriptRoot/.fallout/temp"
 
-$DotNetGlobalFile = "$PSScriptRoot\global.json"
+$DotNetGlobalFile = "$PSScriptRoot/global.json"
 $DotNetInstallUrl = "https://dot.net/v1/dotnet-install.ps1"
 $DotNetChannel = "STS"
 
--- a/src/Fallout.Cli/templates/build.ps1
+++ b/src/Fallout.Cli/templates/build.ps1
@@ -13,9 +13,10 @@ $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent
 # CONFIGURATION
 ###########################################################################
 
-$TempDirectory = Join-Path $PSScriptRoot '_ROOT_DIRECTORY_/.fallout/temp'
+$BuildProjectFile = "$PSScriptRoot\_BUILD_DIRECTORY_\_BUILD_PROJECT_NAME_.csproj"
+$TempDirectory = "$PSScriptRoot\\.fallout\temp"
 
-$DotNetGlobalFile = Join-Path $PSScriptRoot '_ROOT_DIRECTORY_/global.json'
+$DotNetGlobalFile = "$PSScriptRoot\global.json"
 $DotNetInstallUrl = "https://dot.net/v1/dotnet-install.ps1"
 $DotNetChannel = "STS"
 
--- a/src/Fallout.Cli/templates/build.ps1
+++ b/src/Fallout.Cli/templates/build.ps1
@@ -13,10 +13,10 @@ $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent
 # CONFIGURATION
 ###########################################################################
 
-$BuildProjectFile = "$PSScriptRoot\_BUILD_DIRECTORY_\_BUILD_PROJECT_NAME_.csproj"
-$TempDirectory = "$PSScriptRoot\\.fallout\temp"
+$BuildProjectFile = "$PSScriptRoot/_BUILD_DIRECTORY_/_BUILD_PROJECT_NAME_.csproj"
+$TempDirectory = "$PSScriptRoot/.fallout/temp"
 
-$DotNetGlobalFile = "$PSScriptRoot\global.json"
+$DotNetGlobalFile = "$PSScriptRoot/global.json"
 $DotNetInstallUrl = "https://dot.net/v1/dotnet-install.ps1"
 $DotNetChannel = "STS"
 
--- a/src/Fallout.Cli/Commands/SetupCommand.cs
+++ b/src/Fallout.Cli/Commands/SetupCommand.cs
@@ -156,6 +156,8 @@ internal sealed class SetupCommand : IFalloutCommand
                     new
                     {
                         RootDirectory = buildDirectory.GetWinRelativePathTo(rootDirectory),
+                        BuildDirectory = buildProjectRelativeDirectory,
+                        BuildProjectName = buildProjectName,
                         ScriptDirectory = buildDirectory.GetWinRelativePathTo(WorkingDirectory),
                         TargetFramework = TARGET_FRAMEWORK,
                         FalloutVersion = falloutVersion,


ChrisonSimtian (approved):
ah this old chestnut ... io ops in ps1 is a massive pain, thanks for fixing this

ITaluone:
> ah this old chestnut 

Why old? I reverted to the actually "old" ;)

ChrisonSimtian:
ah we use psake at work and the amount of times i fixed this is countless 😂🙈 thats why its an old cheatnut for me. I find this and other things in powershell very brittle, the only advantage over bash is it can natively use dotnet code

ITaluone:
I hate powershell as well :)

For example: simple tools like curl regularly freaking me out, because it is a pain on powershell (which is the "default" terminal on win11). Using good old cmd works like a charm :D

ITaluone:
Ah.. just read the commit message for 7a30bc5.

I try if this is a problem again.

Edit: I tried with forward slashes on windows, for PS v5.1 and PS v7.6.6.. Both worked, and could handle the forward slashes.
I assume this works on unix, too. (I actually cannot imagine why using pwsh on unix since bash is better.. but ok..)

src/Fallout.Cli/templates/build.ps1
```diff
@@ -13,9 +13,10 @@ $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent
 # CONFIGURATION
 ###########################################################################
 
-$TempDirectory = Join-Path $PSScriptRoot '_ROOT_DIRECTORY_/.fallout/temp'
+$BuildProjectFile = "$PSScriptRoot\_BUILD_DIRECTORY_\_BUILD_PROJECT_NAME_.csproj"
+$TempDirectory = "$PSScriptRoot\\.fallout\temp"
```
ITaluone:
```suggestion
$BuildProjectFile = "$PSScriptRoot/_BUILD_DIRECTORY_/_BUILD_PROJECT_NAME_.csproj"
$TempDirectory = "$PSScriptRoot/.fallout/temp"
```

src/Fallout.Cli/templates/build.ps1
```diff
@@ -13,9 +13,10 @@ $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent
...
 
-$DotNetGlobalFile = Join-Path $PSScriptRoot '_ROOT_DIRECTORY_/global.json'
+$DotNetGlobalFile = "$PSScriptRoot\global.json"
```
ITaluone:
```suggestion
$DotNetGlobalFile = "$PSScriptRoot/global.json"
```

build.ps1
```diff
@@ -13,9 +13,9 @@ $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent
 # CONFIGURATION
 ###########################################################################
 
-$TempDirectory = Join-Path $PSScriptRoot '.fallout/temp'
+$TempDirectory = "$PSScriptRoot\\.fallout\temp"
```
ITaluone:
```suggestion
$TempDirectory = "$PSScriptRoot/.fallout/temp"
```

build.ps1
```diff
@@ -13,9 +13,9 @@ $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent
...
 
-$DotNetGlobalFile = Join-Path $PSScriptRoot 'global.json'
+$DotNetGlobalFile = "$PSScriptRoot\global.json"
```
ITaluone:
```suggestion
$DotNetGlobalFile = "$PSScriptRoot/global.json"
```

ChrisonSimtian:
> Ah.. just read the commit message for [7a30bc5](https://github.com/Fallout-build/Fallout/commit/7a30bc58f53a7ee72fb8790cb20e547bf4e8996b).

gotcha, might just have to try and tell ps1 to use the latest v7 not the legacy v5
iirc then MS actually wants to retire it (same as net framework) but keep it semi-alive for legacy reasons.

> which is the "default" terminal on win11
man I feel so much better since I use a macbook as my personal device and run WSL on my win11 work laptop (thanks to claude cli being crap on windows pty :-) ) I even run a LXC on my homelab just for coding. But then energy prices in NZ are not even half of what I used to pay in Germany 10 years ago (and those went up by now by what factor 2?)

ITaluone:
@ChrisonSimtian Why the powershell `$BuildProjectFile = "$PSScriptRoot/build/_build.csproj"` is missing in this repo completely? This is needed for the `:add-package` command.

ChrisonSimtian:
> @ChrisonSimtian Why the powershell `$BuildProjectFile = "$PSScriptRoot/build/_build.csproj"` is missing in this repo completely? This is needed for the `:add-package` command.

`$BuildProjectFile = "$PSScriptRoot/_BUILD_DIRECTORY_/_BUILD_PROJECT_NAME_.csproj"`
isnt that what you're adding here?

ITaluone:
No? that's the template.

I am adding this there, because consumer need it.

I asked why _this_ repo doesn't have this in the `.\build.ps1` (aka was removed intentionally?)

Edit: PR [#204](https://github.com/Fallout-build/Fallout/pull/204) removed this, from the template as well as from the "own" `.\build.ps1`

ChrisonSimtian:
> Edit: PR [#204](https://github.com/Fallout-build/Fallout/pull/204) removed this, from the template as well as from the "own" `.\build.ps1`

correct, and replaced it with the cli call instead in an attempt to make the powershell script as thin as possible.
the linux version is the same btw.

I dont mind you adding it back if its needed but ultimately i'd really like to move away from .ps1 and .sh towards a pure dotnet cli. So its probably better to build that in natively rather than relying on ps1/sh to pass things through

ITaluone:
Ok then.. should this be a follow-up? or should I change the PR to not depend on the bootstrapper files at all?

Your call @ChrisonSimtian 

Edit: The problem is, that the user can name the build project file and folder like he wants. We only give a suggestions as default.
So, as long as the user can choose own names we have to write this names somewhere down. And having yet another file for this mapping, I cannot see the advantage compared to what used to work.

ITaluone:
Ok.. think I have a solution to this:

A `Fallout` build project _always_ have this custom MSBuild variables set like: `<FalloutRootDirectory>` or `<FalloutScriptDirectory>`.

I will change the `AddPackageCommand` to basically make a `grep -R FalloutRootDirectory *.csproj`. If you want, I can introduce a dedicated variable like `<IsFalloutBuildProject>` or similar.

src/Fallout.Cli/templates/build.ps1
```diff
@@ -13,9 +13,9 @@ $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent
 # CONFIGURATION
 ###########################################################################
 
-$TempDirectory = Join-Path $PSScriptRoot '_ROOT_DIRECTORY_/.fallout/temp'
+$TempDirectory = "$PSScriptRoot\\.fallout\temp"
```
ITaluone:
```suggestion
$TempDirectory = "$PSScriptRoot/.fallout/temp"
```

src/Fallout.Cli/templates/build.ps1
```diff
@@ -13,9 +13,9 @@ $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent
...
 
-$DotNetGlobalFile = Join-Path $PSScriptRoot '_ROOT_DIRECTORY_/global.json'
+$DotNetGlobalFile = "$PSScriptRoot\global.json"
```
ITaluone:
```suggestion
$DotNetGlobalFile = "$PSScriptRoot/global.json"
```

build.ps1
ITaluone:
```suggestion
$TempDirectory = "$PSScriptRoot/.fallout/temp"
```

build.ps1
ITaluone:
```suggestion
$DotNetGlobalFile = "$PSScriptRoot/global.json"
```

ChrisonSimtian:
> Ok.. think I have a solution to this:
> 
> 
> 
> A `Fallout` build project _always_ have this custom MSBuild variables set like: `<FalloutRootDirectory>` or `<FalloutScriptDirectory>`.
> 
> 
> 
> I will change the `AddPackageCommand` to basically make a `grep -R FalloutRootDirectory *.csproj`. If you want, I can introduce a dedicated variable like `<IsFalloutBuildProject>` or similar.

lets make a gh issue for this and discuss there independently from this PR. Not in the right headspace for this at the moment, sorry

